Issue #1, opened 2022-05-22
logging module


giovp:
add logging module in order to remove scanpy from requirements.
https://github.com/Delgan/loguru worth giving it a try. 
improve and add logging messages throughout repo

rahulbshrestha:
Fixed by #17

PR #17, opened 2022-07-27
Loguru logging module

rahulbshrestha:
This PR replaces the logging function from scanpy with [loguru](https://github.com/Delgan/loguru).

A couple of logging messages (and tests) have been added throughout the repo.

Related issue: https://github.com/scverse/napari-spatialdata/issues/1


rahulbshrestha:
Remaining tasks:
- Add loguru to dependencies
- Add test to simulate: setting shapes and exporting with "shift-e"

PR #20, opened 2022-08-05
multicolor text

tothmarcella:


src/napari_spatialdata/_utils.py
```diff
@@ -99,20 +99,24 @@ def _get_categorical(
...
     kdtree = KDTree(coords)
     clusters = np.full(len(coords), fill_value="", dtype=object)
     # index consists of the categories that need not be string
     clusters[kdtree.query(df.values)[1]] = df.index.astype(str)
+    order = np.argsort(kdtree.query(df.values)[1])
+    colortypes = colortypes.reindex(order)
+    colortypes = np.array([c[0] for c in colortypes])
     # napari v0.4.9 - properties must be 1-D in napari/layers/points/points.py:581
     colors = np.array([to_hex(col) for col in colors])
     colors = np.array([col if not len(cl) else to_hex((0, 0, 0)) for cl, col in zip(clusters, colors)])
```
tothmarcella, line 118: If I saw correctly `"colors"` was never used even before I changed anything. if that is really the case then maybe it would be useful to delete this part of the code and as an input of the function have the color pallette instead of the color vector of the points? (but maybe too many things would have to be changed)

giovp:
@tothmarcella please install pre-commit and run pre-commit on the files, check [here](https://github.com/scverse/squidpy_notebooks/blob/main/CONTRIBUTING.rst) how to do it, thank you!

tothmarcella:
added a [new topic](https://forum.image.sc/t/adding-multi-colored-text-to-clusters-of-points/70451?u=tothmarcella) on image.sc

src/napari_spatialdata/_utils.py
giovp: I think they are already removed in main

giovp:
hi @tothmarcella ! what's the status of this PR? I think napari 0.4.17rc might not be too far, we could merge this right after that.

tothmarcella:
Hi @giovp it is done but I dont think it was reviewed yet

src/napari_spatialdata/_model.py
```diff
@@ -66,7 +66,7 @@ def get_items(self, attr: str) -> Tuple[str, ...]:
             if self.points_var is not None:
                 return tuple(map(str, self.points_var.unique()))
             else:
-                return tuple(["No points found."])  # noqa: C409
+                return tuple(["No points found."])
         return tuple(map(str, getattr(self.adata, attr).index))
```
giovp, line 70: I feel `# noqa: C409` needs to be re added to the return otherwise it'd fail again.

giovp (commented):
great job! only minor comments and then we can merge!

giovp:
also you might need to re-merge from main to fix the pre-commit check

giovp:
fantastic job @tothmarcella ! 🎉 LGTM!

Issue #3, opened 2022-06-22
scatterplot to visualize genes and annotation in `adata.obsm` coordinates


giovp:
goal is to have a way to visualize points in 2D coordinates (e.g. umap, pca, spatial scatterplot) with gene expression and annotation. 
Ideally, we won't re-invent the wheel but use what's already out there, possible options:
-  https://github.com/matplotlib/napari-matplotlib
- https://github.com/lukasz-migas/napari-plot
  - in particular this example: https://github.com/lukasz-migas/napari-plot/blob/main/examples/plot-scatter.py

if would be nice to have some type of interactive selection of the points as well

PR #26, opened 2022-09-19
Scatterplot Widget

rahulbshrestha:
This PR introduces a scatterplot widget to napari-spatialdata, which is able to visualise points in 2D coordinates.

Related issue: https://github.com/scverse/napari-spatialdata/issues/3
Contributors: @rahulbshrestha @tothmarcella @ganesh25


src/napari_spatialdata/_widgets.py
```diff
@@ -232,6 +240,65 @@ def _(self, vec: pd.Series, key: str, layer: Layer) -> dict[str, Any]:
...
+            except Exception as e:  # noqa: B902
+                logger.error(e)
+                continue
+            self.data = vec
+        print("Selected data: ", self.data)
+        return
+
+    def getData(self) -> NDArrayA:
+        # check type?
+        return self.data
```
giovp, line 265: use property and setter as in self.text (see below)

src/napari_spatialdata/_widgets.py
```diff
@@ -232,6 +240,65 @@ def _(self, vec: pd.Series, key: str, layer: Layer) -> dict[str, Any]:
...
+            raise ValueError(f"{field} is not a valid adata field.")
+        self._attr = field
+        self._getter = getattr(self.model, f"get_{field}")
+        self.attrChanged.emit()
+
+    def getAttribute(self) -> Optional[str]:
+        if TYPE_CHECKING:
+            assert isinstance(self._attr, str)
+        return self._attr
+
```
giovp, line 280: use property and setter as in self.text (see below)

src/napari_spatialdata/_widgets.py
```diff
@@ -232,6 +240,65 @@ def _(self, vec: pd.Series, key: str, layer: Layer) -> dict[str, Any]:
...
+    def setAttribute(self, field: Optional[str]) -> None:
+        if field == self.getAttribute():
+            return
+        if field not in ("var", "obs", "obsm"):
+            raise ValueError(f"{field} is not a valid adata field.")
+        self._attr = field
+        self._getter = getattr(self.model, f"get_{field}")
+        self.attrChanged.emit()
+
```
giovp, line 275: can we use setters and getters also here?

src/napari_spatialdata/_widgets.py
```diff
@@ -232,6 +240,65 @@ def _(self, vec: pd.Series, key: str, layer: Layer) -> dict[str, Any]:
...
+        self.text = text  # type: ignore[assignment]
+
+        if self.getAttribute() == "var":
+            if TYPE_CHECKING:
+                assert isinstance(text, str)
+            super().setAdataLayer(text)
+        elif self.getAttribute() == "obsm":
+            if TYPE_CHECKING:
+                assert isinstance(text, int)
+            super().setIndex(text)
```
giovp, line 291: could this method be part of AListWidget natively? instead of subclassing?

src/napari_spatialdata/_view.py
```diff
@@ -19,12 +19,203 @@
...
+
+        # Matplotlib
+
+        self.matplotlib_widget = MatplotlibWidget(self.viewer, self.model)
+        self.layout().addWidget(self.matplotlib_widget)
+
+        # Dropdown menu to select between obs, obsm, var for X axis
+        x_selection_label = QLabel("Select type for X axis:")
+        x_selection_label.setToolTip("Select between obs, obsm and var.")
+        self.x_selection_widget = QComboBox()
```
giovp, line 63: we discussed about ordering, is QComboBox still the preferred solution or is there an alterantive to use space better in the widget?

src/napari_spatialdata/_view.py
```diff
@@ -19,12 +19,203 @@
...
+        self.layout().addWidget(self.y_widget)
+        self.layout().addWidget(self.y_component_widget)
+
+        self.y_selection_widget.currentTextChanged.connect(self.y_widget.setAttribute)
+        self.y_selection_widget.currentTextChanged.connect(self.y_component_widget.setAttribute)
+        self.y_selection_widget.currentTextChanged.connect(self.y_component_widget.setToolTip)
+
+        # Color
+        color_selection_label = QLabel("Select type for color:")
+        color_selection_label.setToolTip("Select between obs and var.")
```
giovp, line 124: there is a lot of code repetition here. Is there maybe a way to put this in a widget itself? like https://github.com/scverse/napari-spatialdata/blob/fccc67a08b5f5073276e9f9fd03937daa6d8e843/src/napari_spatialdata/_widgets.py#L302

src/napari_spatialdata/_view.py
```diff
@@ -19,12 +19,203 @@
...
+        )
+        if "points" in layer.metadata:
+            self.model.points_coordinates = layer.metadata["points"].X
+            self.model.points_var = layer.metadata["points"].obs["gene"]
+            self.model.point_diameter = np.array([0.0] + [layer.metadata["point_diameter"]] * 2) * self.model.scale
+        self.model.spot_diameter = (
+            np.array([0.0] + [Key.uns.spot_diameter(self.model.adata, Key.obsm.spatial, self.model.library_id)] * 2)
+            * self.model.scale
+        )
+        self.model.labels_key = layer.metadata["labels_key"] if isinstance(layer, Labels) else None
```
giovp, line 192: why would you need this? I don't think you need to plot the stuff in the scatter back in the spatial coordinate right? better to clean up

setup.cfg
```diff
@@ -44,6 +44,7 @@ install_requires =
     squidpy
     napari[all] @ git+https://github.com/napari/napari.git#egg=napari[all]
     loguru
+    napari-matplotlib
```
giovp, line 47: is there a way to have napari-matplotlib as a dependency, but not showing it up in the plugin bar? 
image

it's no problem, but I don't think the user should interact with it (as it interacts via our interface)/

src/napari_spatialdata/_widgets.py
```diff
@@ -232,6 +240,65 @@ def _(self, vec: pd.Series, key: str, layer: Layer) -> dict[str, Any]:
...
+
+    def _onAction(self, items: Iterable[str]) -> None:
+        for item in sorted(set(items)):
+            try:
+                vec, _ = self._getter(item, index=self.getIndex())
+            except Exception as e:  # noqa: B902
+                logger.error(e)
+                continue
+            self.data = vec
+        print("Selected data: ", self.data)
```
giovp, line 260: don't use prints but proper logging

src/napari_spatialdata/_widgets.py
```diff
@@ -232,6 +240,65 @@ def _(self, vec: pd.Series, key: str, layer: Layer) -> dict[str, Any]:
...
 
 
+class ScatterListWidget(AListWidget):
+    attrChanged = Signal()
+    _text = None
+
+    def __init__(self, viewer: Viewer, model: ImageModel, attr: str, **kwargs: Any):
+        AListWidget.__init__(self, viewer, model, attr, **kwargs)
+        self.attrChanged.connect(self._onChange)
+        self.data = -1
```
giovp, line 250: what's this? don't set default unless really necessary, and if necessary set them meaningful or add comments.

src/napari_spatialdata/_widgets.py
```diff
@@ -232,6 +240,65 @@ def _(self, vec: pd.Series, key: str, layer: Layer) -> dict[str, Any]:
...
+
+    def _onAction(self, items: Iterable[str]) -> None:
+        for item in sorted(set(items)):
+            try:
+                vec, _ = self._getter(item, index=self.getIndex())
+            except Exception as e:  # noqa: B902
+                logger.error(e)
+                continue
+            self.data = vec
```
giovp, line 259: I think the attribute `.data` should be either set as a property for proper getter/setter or should not be used and the vec should just be returned

src/napari_spatialdata/_widgets.py
```diff
@@ -232,6 +240,65 @@ def _(self, vec: pd.Series, key: str, layer: Layer) -> dict[str, Any]:
...
+class ScatterListWidget(AListWidget):
+    attrChanged = Signal()
+    _text = None
+
+    def __init__(self, viewer: Viewer, model: ImageModel, attr: str, **kwargs: Any):
+        AListWidget.__init__(self, viewer, model, attr, **kwargs)
+        self.attrChanged.connect(self._onChange)
+        self.data = -1
+
+    def _onAction(self, items: Iterable[str]) -> None:
```
giovp, line 252: this is the key action of the widget, which instead of creating a new napari layer (with points/labels), it returns the vec for the napari matplotlib, correct? Here you chosed to add this as a method after subclassing AListWidget, to recapitulate the options are:

1. change the `onAction` method	logic of `AListWidget` directly, to make the method behave differently according to some arguemnt (e.g. scatter v. view).
2. add this method directly to `AListWidget` but as a separate method called `_onActionScatter`
3. the way you are doing it now.

Considering that maybe (from first glance) the `ScatterListWidget` could actually be trimmed down to only this method, I think it's worth discussing what's the best option. I think all the 3 are reasonable (yet have different pro and cons). Curious to hear what you end up proposing!

src/napari_spatialdata/_widgets.py
giovp: I mean both for `setAttribute` and `getAttribute`

src/napari_spatialdata/_widgets.py
```diff
@@ -260,6 +327,53 @@ def addItems(self, texts: Union[QtWidgets.QListWidgetItem, int, Iterable[str]])
         super().addItems(tuple(texts))
 
 
+class ComponentWidget(ObsmIndexWidget):
```
giovp, line 330: same comments for the above class also here

src/napari_spatialdata/_widgets.py
```diff
@@ -379,6 +493,34 @@ def cmap(self) -> str:
         return self._cmap
 
 
+class MatplotlibWidget(NapariMPLWidget):
+    def __init__(self, viewer: Viewer, model: ImageModel):
+
+        super().__init__(viewer)
+
+        self.viewer = viewer
+        self.model = model
```
giovp, line 502: how is self.model used in the other widgets, is it `_model` or `model` ? Please align all of them, and same for viewer.

src/napari_spatialdata/_widgets.py
```diff
@@ -379,6 +493,34 @@ def cmap(self) -> str:
...
+        self.axes = self.canvas.figure.subplots()
+
+    def clear(self) -> None:
+        self.axes.clear()
```
giovp, line 506: general design principle, put methods/attributes from the most complex to the least complex in a class. This one for isntance it's just one line of code, so should be put at the end of the class, whereas more complex methods (either in terms of quantity of code, or diffficulty of logic) should be put on top.

giovp (changes requested):
it looks already really good, left few initial comments. Please check out also the minor changes I did in the previous commit.

giovp:
@rahulbshrestha there is also a test for the logger failing, cool that it's catching this!

src/napari_spatialdata/_widgets.py
rahulbshrestha: It varies among the widgets. For example, in class ListWidget self._viewer is used and in class AListWidget self.viewer is used. Should I change all of them to the first example i.e `self._viewer = viewer`.

giovp: yes please, feel free to do it everywhere if appropriate. Maybe add it as a property

src/napari_spatialdata/_widgets.py
tothmarcella: because of the way onAction is connected to different events I think it would be easier to set data as property (I think it would be hard to handle a return in some cases) so I will do that then if thats ok

src/napari_spatialdata/_widgets.py
tothmarcella: ok. I wrote these based on setindex and getindex functions in the alistwdget class for self._index, should it also be changed there for property setter and getter or is it a different situation?

src/napari_spatialdata/_widgets.py
tothmarcella: i will set as default to None. the idea was to use a default to check if anything has been selected before attempting to do the scatterplot

src/napari_spatialdata/_widgets.py
giovp: good point, because of the attribute getter and emit maybe ok to not have as property

src/napari_spatialdata/_view.py
rahulbshrestha: We switched from QVBoxLayout to QGridLayout -> the UI looks better now.

setup.cfg
rahulbshrestha: Kevin (@kevinyamauchi) mentioned that this isn't fixable right now but we can create an issue in the napari github repo. (I'll create one there)

docs/conf.py
```diff
@@ -46,7 +46,7 @@
 
 # General information about the project.
 project = "napari-spatialdata"
-copyright = "2022, giovanni palla"  # noqa: A001
+copyright = "2022, giovanni palla"
 author = "giovanni palla"
```
giovp, line 50:
```suggestion
author = "scverse team"
```

giovp (approved):

giovp:
I'd like to quickly try it out but can't rn, will do after meeting. After that, I'll merge!

giovp:
trying it out, super nice, color renders really nicely. 

One thing, when categorical is chosen, the colormap is not updated (should be removed) and the legend is not visible:

image

with continuous colors though works well!
one thing, it'd be nice to put the color variables (e.g. obs: row_num here) in the title of the plot, so that the info for the variable used to color is also present (the x and y axis labels looks great btw)

image

giovp:
another funny thing I noticed is that as I continue changing variables for axis and colors, the scatterplot keeps getting narrower to a point were the dots can't be distinguished anymore. Really weird not sure what could cause it
image

giovp:
finally, not sure I mentioned before but this bit for getting the colors that reassign the `vec` value to anndata should be changed
https://github.com/scverse/napari-spatialdata/blob/aec224238a3a9e9df93f7fb977f6a838262bedd2/src/napari_spatialdata/_utils.py#L86-L89

my understanding is that it is needed in order to then assign the categorical color in un in the following function call. 
If that's the case, then couldn't just the vector be passed to that function and the colormap could be then just generated from it? In other words, existing anndata entries should never be modified in place (or even as copy) by the widgets. Let's discuss today alternatives to current behaviour.

giovp:
@rahulbshrestha @tothmarcella I'll go ahead and merge this. There is a fair bit of refactoring in handling categoricals that we will have to do but for another PR. 

Great work!

PR #32, opened 2022-12-17
add docs

giovp:
docs of the widget looks really ugly (tons of attributes and methods unused) need to investigate better.


giovp:
ok so there is practically no sane way to render pyqt-inherited classes with sphinx, unless with many disabled warning and nitpicks. Not sure whether we want to have the widgets documented at all maybe examples and higher level interface to spatialdata will be enough.

PR #44, opened 2023-03-08
Update layers in-place

rahulbshrestha:
This PR introduces in place layer updates to napari-spatialdata. Clicking on obsm, obs, or var after selecting a layer replaces the already existing layer, rather than creating a new one.


LucaMarconato:
Looking forward to use this Rahul! One note about performance. I see that currently the old layer is deleted and a new one is created. I wonder if changing the color of the existing shapes, without having to recreate them, would be more performant, or if it would be the same. Asking an opinion also to @kevinyamauchi

PR #47, opened 2023-03-22
update install and precommit

giovp:


giovp:
@rahulbshrestha @tothmarcella it worked, will merge to main and then you can rebase from here

PR #58, opened 2023-04-27
Fixing private module imports

rahulbshrestha:
Replacing `napari_spatialdata._interactive import Interactive` with `from napari_spatialdata import Interactive`

### DIFF
--- a/src/napari_spatialdata/__init__.py
+++ b/src/napari_spatialdata/__init__.py
@@ -12,6 +12,7 @@ except ImportError:
 
 del version, parse
 
+from napari_spatialdata._interactive import Interactive as Interactive  # noqa: E402
 from napari_spatialdata._view import (  # noqa: E402
     QtAdataScatterWidget,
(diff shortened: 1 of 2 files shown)


src/napari_spatialdata/__init__.py
```diff
@@ -12,6 +12,7 @@
 
 del version, parse
 
+from napari_spatialdata._interactive import Interactive as Interactive  # noqa: E402
```
giovp, line 15: thanks for this! why the renaming Interactive as Interactive and not just from ... import Interactive?

giovp (approved):

src/napari_spatialdata/__init__.py
rahulbshrestha: mypy threw an error: `error: Module "napari_spatialdata" does not explicitly export attribute "Interactive"  [attr-defined]`. Can be fixed this way or with `__all__ = ["Interactive"]`

PR #61, opened 2023-05-03
adding transformation

tothmarcella:


giovp (approved):
looks great! hav eyou tested that it works on the merfish example we had issues in the hackathon?

tothmarcella:
> looks great! hav eyou tested that it works on the merfish example we had issues in the hackathon?

yes it works @giovp

Issue #66, opened 2023-05-04
RBG images are shown as multichannel images


LucaMarconato:
Images should be rendered with `rgb=True` when they have 3 channels.

kevinyamauchi:
@giovp , @rahulbshrestha , @tothmarcella , would one of you be able to work on this?

melonora:
I already have code for this too:) it was in #93 which was mostly a port from Luca's code on the spatialdata branch, but then with some utility functions.

giovp:
@rahulbshrestha should have taken that and ported to another PR @rahulbshrestha could you comment here?

rahulbshrestha:
yup, I put up a PR here based on Wouter's code: https://github.com/scverse/napari-spatialdata/pull/115

melonora:
Completed with #115

PR #93, opened 2023-05-23
Refactor widgets and layer metadata

### LINKED ISSUE #55: Set a reasonable upper bound of points that we want to render
napari currently can't render more than X points. Find a reasonable value for X (probably something between 100K and 1M) and change the code to subset to this number of points (load only those in-memory from dask and show only them).

### LINKED ISSUE #87: Add element to layer metadata
On main, currently the layer does not know which SpatialElement it is visualizing. This makes it impossible to change visibility of the individual layers when a coordinate system is selected in which the spatial element is not present.

melonora:
This PR moves the widgets in interactive to a script specific for data widgets. On top of that it merges the code from @LucaMarconato regarding adding metadata to the layers with the current code on main. 

Utility functions have been moved to utils. Performance has been improved by submitting points and polygons data based on the maximum amount to be visualized, before performing any other calculations.

Also a bug has been fixed for the point data (x and y coordinates were not swapped, napari expects yx).

Closes #87 and #66, potentially closes #55


melonora:
@rahulbshrestha @tothmarcella Hi both, Could you please have a look at the tests that are failing and the code in general? It seems to be due to a missing attribute and a mismatch in dimensions in anndata.obs and blobs.shape. 

I think the changes are following what is intended so it could be that we have to change some tests.

rahulbshrestha:
Hi @melonora! Yup, I'll take a look.

melonora:
For the regionprops, we could also use dask regionprops. Would there be any objection to that?

src/napari_spatialdata/_utils.py
```diff
@@ -221,3 +203,272 @@ def _points_inside_triangles(points: NDArrayA, triangles: NDArrayA) -> NDArrayA:
...
+    # merged.loc[index_of_background, "v"] = 0
+    merged["v"] = merged["v"].astype(int)
+
+    assert len(annotating_rows) == len(merged)
+    assert annotating_rows.obs[instance_key].tolist() == merged["v"].tolist()
+
+    merged_centroids = merged[["cx", "cy"]].to_numpy()
+    assert len(merged_centroids) == len(merged)
+    annotating_rows.obsm["spatial"] = merged_centroids
+    annotating_rows.obsm["region_radius"] = np.array([10.0] * len(merged_centroids))  # arbitrary value
```
melonora, line 339: This part is due to squidpy legacy code. Can this also be removed?

src/napari_spatialdata/_utils.py
```diff
@@ -221,3 +203,272 @@ def _points_inside_triangles(points: NDArrayA, triangles: NDArrayA) -> NDArrayA:
...
+            assert len(v) == 1
+            xdata = v.__iter__().__next__()
+            list_of_xdata.append(xdata)
+        new_raster = list_of_xdata
+
+    return new_raster, rgb
+
+
+def _init_colors_for_obs(adata: AnnData) -> AnnData:
+    from scanpy.plotting._utils import _set_colors_for_categorical_obs
```
giovp, line 263: this we copied over it frm scanpy so to not be dependent on scanpy anymore, it's in `categorical_utils.py`

setup.cfg
```diff
@@ -45,6 +45,7 @@ install_requires =
     napari-matplotlib
     scikit-learn
     spatialdata
+    scanpy
```
giovp, line 48: remove scanpy  as requirement

src/napari_spatialdata/_utils.py
```diff
@@ -221,3 +203,272 @@ def _points_inside_triangles(points: NDArrayA, triangles: NDArrayA) -> NDArrayA:
...
+
+
+def _get_mapping_info(annotation_table: AnnData) -> tuple[str, str, str]:
+    regions = annotation_table.uns["spatialdata_attrs"]["region"]
+    regions_key = annotation_table.uns["spatialdata_attrs"]["region_key"]
+    instance_key = annotation_table.uns["spatialdata_attrs"]["instance_key"]
+    return regions, regions_key, instance_key
+
+
+def _find_annotation_for_labels(
```
giovp, line 290: it's unclear why this is needed, in the implementation from main there shouldn't be need for this. Is it possible that this is something you ported over from luca's branch and now it's here for legacy reason?

src/napari_spatialdata/_interactive.py
```diff
@@ -1,203 +1,12 @@
...
-            self._sdata.labels[key],
-            name=key,
-            affine=affine,
-            metadata={
-                "adata": self._sdata.table[
-                    self._sdata.table.obs[self._sdata.table.uns["spatialdata_attrs"]["region_key"]] == key
-                ],
-                "labels_key": self._sdata.table.uns["spatialdata_attrs"]["instance_key"],
-            },
-        )
```
giovp, line 167: this part is replaced by the big/complex function below. I think this way to handle annotation metadata is much cleaner and would keep this.

src/napari_spatialdata/_utils.py
```diff
@@ -221,3 +203,272 @@ def _points_inside_triangles(points: NDArrayA, triangles: NDArrayA) -> NDArrayA:
...
+    assert annotating_rows.obs[instance_key].tolist() == merged["v"].tolist()
+
+    merged_centroids = merged[["cx", "cy"]].to_numpy()
+    assert len(merged_centroids) == len(merged)
+    annotating_rows.obsm["spatial"] = merged_centroids
+    annotating_rows.obsm["region_radius"] = np.array([10.0] * len(merged_centroids))  # arbitrary value
+    return annotating_rows
+
+
+def _find_annotation_for_shapes(
```
giovp, line 343: same here as before

src/napari_spatialdata/_utils.py
```diff
@@ -221,3 +203,272 @@ def _points_inside_triangles(points: NDArrayA, triangles: NDArrayA) -> NDArrayA:
...
+    annotating_rows.uns[Key.uns.spatial][element_path] = {}
+    annotating_rows.uns[Key.uns.spatial][element_path][Key.uns.scalefactor_key] = {}
+    annotating_rows.uns[Key.uns.spatial][element_path][Key.uns.scalefactor_key]["tissue_hires_scalef"] = 1.0
+    annotating_rows.obsm["spatial"] = spatial
+    # workaround for the legacy code to support different sizes for different circles
+    annotating_rows.obsm["region_radius"] = radii
+    return annotating_rows
+
+
+def _find_annotation_for_regions(
```
giovp, line 399: same here

src/napari_spatialdata/_utils.py
```diff
@@ -221,3 +203,272 @@ def _points_inside_triangles(points: NDArrayA, triangles: NDArrayA) -> NDArrayA:
...
+            else:
+                raise TypeError(f"Unsupported shape type: {type(shape)}")
+        else:
+            raise ValueError(f"Unsupported element type: {type(base_element)}")
+        return table
+
+    return None
+
+
+def _get_ellipses_from_circles(centroids: NDArrayA, radii: NDArrayA) -> NDArrayA:
```
giovp, line 438: same here

giovp (changes requested):
I would revert back to what we had before regarding passing annotation metadata to layers and to add layers. imho the spatialdata widget should be kept the same as we had before. @tothmarcella @rahulbshrestha could you also give it a comment here thanks!

src/napari_spatialdata/_sdata_widgets.py
```diff
@@ -0,0 +1,261 @@
...
+                f"being correct beyond a certain zoom level"
+            )
+            # TODO: If https://github.com/scverse/napari-spatialdata/issues/35 is fixed, use points for circles, faster
+            self._viewer.add_points(
+                coords_yx,
+                name=key,
+                size=2 * radii,
+                metadata=metadata,
+                affine=affine,
+            )
```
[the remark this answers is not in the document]
LucaMarconato: In `main` currently only ellipses are used, this is bad for performance and user experience, so definitely we should prefer always points to always shapes, I will make a PR. Re always points vs always shapes, I'll experiment a bit with small Visium datasets and get back to it.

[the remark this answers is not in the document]
melonora: Note that there is work ongoing in napari at the moment to actually fix this issue. Might already be included in 0.4.18 which is aimed for release next month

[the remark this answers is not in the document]
LucaMarconato: Very nice!

melonora:
Closing this in favor of new PR

PR #99, opened 2023-06-02
Layer cs visibility

melonora:
This PR is the replacement for #96. In its current state it has the following behavior when having multiple coordinate spaces:

Layer.metadata has a field to keep track of the visibility state in a coordinate system.

When starting napari
1. When a layer is present in the selected coordinate system and is made visible `active_in_cs` metadata is a set containing only the name of the coordinate system.
   If made invisible afterwards, the selected coordinate system is removed from the set.
2. If present and visible, but then changing to a different coordinate system in which the spatialdata element is not present, `layer.visible` is set to `False`, but `active_in_cs` still contains the name of the coordinate system in which it was made visibe. If present but not visible and then changing, nothing changes. 
3. When going back to the original coordinate system, the `active_in_cs` metadata of the layers visualizing the elements in that coordinate system are checked to whether a layer should become visible again or not.

Even when an element is not present in the selected coordinate system, it can still be made visible again by clicking on the eye. This would allow for now for the landmark registration approach. However, the name of the coordinate system is not added to `active_in_cs` in this case.

Note that changing the visibility does cause the scene to have to be redrawn upon visualizing again, which can take time (image is loaded again).


src/napari_spatialdata/_interactive.py
```diff
@@ -77,7 +80,36 @@ def _onClick(self, text: str) -> None:
...
+        coordinate_space = self.coordinate_system_widget._system
+
+        # No layer selected on first time coordinate system selection
+        if self._viewer.layers:
+            for layer in self._viewer.layers:
+                if layer.name not in elements:
+                    layer.visible = False
+                elif layer.metadata["active_in_cs"]:
+                    layer.visible = True
+                    # Prevent _update_visible_in_cs of removing coordinate system if previously in coordinate system.
```
melonora, line 108: This here ensures that the coordinate system in which a layer is visible is not removed when changing back to a previously selected coordinate system in which the layer is present and was visible. Otherwise in `_update_visible_in_cs`

```
else:
      layer_active.remove(selected_coordinate_system)
```
This would ensure that the coordinate system would stay removed from the metadata, which should not happen.

melonora:
The failing test is caused by a dependence on [#282](https://github.com/scverse/spatialdata/pull/282#event-9414922489) which is not used for tests here yet.

melonora:
@LucaMarconato @giovp If you have time please have a look and let me know if you agree with the applied functionality.

kevinyamauchi:
Hey @melonora , it looks like like the tests are failing because they are expecting to use the version of `blobs()` on `spatialdata` main (introduced in https://github.com/scverse/spatialdata/pull/282), which hasn't been released to pypi yet. I just cut a release. Once you see spatialdata `0.0.10`, please try re-running the tests. Sorry about that!

melonora:
Hi @kevinyamauchi ,
No worries, I knew the reason:) I hope to push the expected PR regarding saving back points / shapes and labels to spatialdata quickly after this one. That PR will be branched from the branch I used here so would be nice to have it in beforehand.

kevinyamauchi:
Thanks for the fix, @melonora . Do you know the source of the docs failure?

melonora:
It does not seem to recognize napari_spatialdata as a module. This also causes the docs builds in the other PRs to fail.

melonora:
I am able to locally build the docs. @giovp @rahulbshrestha any ideas? Are you able to locally build?

src/napari_spatialdata/_interactive.py
```diff
@@ -77,7 +80,36 @@ def _onClick(self, text: str) -> None:
         elif self.elements_widget._elements[text] == "shapes":
             self._add_shapes(text)
 
+    def _update_visible_in_cs(self, event: Event) -> None:
```
kevinyamauchi, line 83: I would prefer if `cs` was written out as `coordinate_system` in the function name. I think this is much more clear.

src/napari_spatialdata/_interactive.py
```diff
@@ -77,7 +80,36 @@ def _onClick(self, text: str) -> None:
...
+        if layer.name in elements:
+            if selected_coordinate_system not in layer_active:
+                layer_active.add(selected_coordinate_system)
+            else:
+                layer_active.remove(selected_coordinate_system)
+
+    def _update_layers_visibility(self) -> None:
+        """Toggle layer visibility dependent on presence in currently selected coordinate system."""
+        elements = self.elements_widget._elements
+        coordinate_space = self.coordinate_system_widget._system
```
kevinyamauchi, line 99: It would be nice if we could unify the usage of "coordinate system" and "coordinate space". I think we call them coordinate systems in SpatialData, but I'm not 100% sure. @LucaMarconato  can you confirm?

src/napari_spatialdata/_interactive.py
```diff
@@ -163,41 +201,44 @@ def _add_label(self, key: str) -> None:
                     self._sdata.table.obs[self._sdata.table.uns["spatialdata_attrs"]["region_key"]] == key
                 ],
                 "labels_key": self._sdata.table.uns["spatialdata_attrs"]["instance_key"],
+                "active_in_cs": {selected_cs},
```
kevinyamauchi, line 204: Do I understand this correctly that we are assuming that whenever "_add_label()" is called, the new labels layer is active? Are there cases where this might not be true? I think we should probably make a note of this in the docstring. This could be a uprising side effect down the road.

(this would hold for the other `_add_*` methods)

tests/test_spatialdata.py
```diff
@@ -100,3 +101,57 @@ def test_sdatawidget_points(caplog, make_napari_viewer: Any):
...
+
+    # Load 2 layers, make 1 invisible
+    widget._onClick(list(sdata.points.keys())[0])
+    widget._onClick(list(sdata.labels.keys())[0])
+
+    points = viewer.layers[0]
+    labels = viewer.layers[1]
+
+    assert points.metadata["active_in_cs"]
+    assert labels.metadata["active_in_cs"]
```
kevinyamauchi, line 123: What is the intention of the test here? To test if they key `"active_in_cs"` is present in the metadata dictionary? If so, I think it should be `assert "active_in_cs" in points.metadata`

src/napari_spatialdata/utils/_test_utils.py
```diff
@@ -0,0 +1,54 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from PyQt6.QtCore import Qt
```
kevinyamauchi, line 5: We shouldn't directly import from PyQt5/6 or pyside. Users should be able to use either Qt bindings. We should instead use qtpy as the compatibilty layer. Please update all Qt-related imports to use `qtpy`

See here for an example: https://github.com/napari/napari/blob/e1415c59ca3a1de78ac7e31784edd46a04155b9e/napari/_qt/layer_controls/qt_image_controls.py#L3-L11

kevinyamauchi: In case you're curious the readme does a good job explaining why we use `qtpy`: https://github.com/spyder-ide/qtpy

kevinyamauchi: Also, do we support Qt6? It isn't even really fully supported in napari yet (cc @giovp )

kevinyamauchi: Is this why you needed to patch the docs for Qt6 in #107

kevinyamauchi (changes requested):
Hey @melonora . Thanks for the PR! I think it's on the right track. The tests were helpful for understanding the intention of the PR.

I have made an initial review with some comments below. The functionality seems to work. The does logic feels a bit convoluted to me. It may be that it's as simple as it can get, but I just want to check in. To understand a bit better, I have a few questions:

- To me it looks like the `active_in_cs` metadata that is tracking which coordinate system a layer can be active in is populated each time a new coordinate system is selected. This feels a bit odd because it's value is changing as the user clicks different coordinate systems even though the underlying ground truth is not. Is it possible for this to be populated upon layer instantiation and then just checked each time a coordinate system is changed? I think this would simplify the logic in places
- If the `active_in_cs` is modified by the user, does the view get updated? Should we use an [EventedSet](https://napari.org/stable/api/napari.utils.events.EventedSet.html) for the `active_in_cs` metadata? That way if that is updated, the view can be updated? We could also save this for a follow up PR
- If a transformation is added, do we need to update the `active_in_cs`metadata?

src/napari_spatialdata/_interactive.py
melonora: agree, long function name though hehe

src/napari_spatialdata/_interactive.py
melonora: At the moment we only directly add layers by using the functions in the sdata widget. These ultimately give a call to viewer.add_<layer_type> for which no parameter `visible` is used now. 

I think what we need and which will be part of a follow up PR (my next one or second next) for also saving shapes / points is a spatialdata viewer which holds a reference to the napari viewer or inherits from it so we can have some additional callbacks. Here we could add an option instead of `add_image` and so on to `add_spatialelement_layer` which is ultimately `add_layer` with some extra bells and whistles.

src/napari_spatialdata/utils/_test_utils.py
melonora: No, the docs `conf.py` stated already that PyQt6 is being used. The doc builds in other PRs were also failing for the same reason as I suspect that PyQt6 dependencies changed. The updated dependency was not supported on Ubuntu 20.04.

melonora: Resolved with [b6953f4](https://github.com/scverse/napari-spatialdata/pull/99/commits/b6953f44a00101034b47a2a01d4121046c988451)

src/napari_spatialdata/_interactive.py
melonora: True we use coordinate systems. Resolved with [2ec5203](https://github.com/scverse/napari-spatialdata/pull/99/commits/2ec5203d9858db9952e7375f5374d042a4e06738)

src/napari_spatialdata/_interactive.py
melonora: With the follow up PR for saving points / shapes I add the current coordinate system metadata to indicate in which coordinate system the layer lives. When then adding a new layer by clicking on one of the napari buttons to add the layer, it checks the metadata of the layer which was active to inherit metadata regarding coordinate system and spatialdata object. 
While technically we could add these custom callbacks in the sdata widget, it feels convoluted to me so that is why I would go for the viewer solution I talked about in the comment above.

src/napari_spatialdata/_interactive.py
melonora: Resolved with [2ec5203](https://github.com/scverse/napari-spatialdata/pull/99/commits/2ec5203d9858db9952e7375f5374d042a4e06738)

melonora:
> Hey @melonora . Thanks for the PR! I think it's on the right track. The tests were helpful for understanding the intention of the PR.
> 
> I have made an initial review with some comments below. The functionality seems to work. The does logic feels a bit convoluted to me. It may be that it's as simple as it can get, but I just want to check in. To understand a bit better, I have a few questions:
> 
>     * To me it looks like the `active_in_cs` metadata that is tracking which coordinate system a layer can be active in is populated each time a new coordinate system is selected. This feels a bit odd because it's value is changing as the user clicks different coordinate systems even though the underlying ground truth is not. Is it possible for this to be populated upon layer instantiation and then just checked each time a coordinate system is changed? I think this would simplify the logic in places
> 
>     * If the `active_in_cs` is modified by the user, does the view get updated? Should we use an [EventedSet](https://napari.org/stable/api/napari.utils.events.EventedSet.html) for the `active_in_cs` metadata? That way if that is updated, the view can be updated? We could also save this for a follow up PR
> 
>     * If a transformation is added, do we need to update the `active_in_cs`metadata?

@kevinyamauchi Thanks for reviewing! I will try to clarify some things:

- Active_in_cs tracks in which coordinate systems the layer is visible / editable. This means that if we have a spatial element having 2 coordinate systems associated with it `global` and `other` and being open as a layer, if visible and switching from `global` to `other` this means `active_in_cs` will change from `("global")` to `("global", "other")`.
- If now toggling visibility of the layer in other so it is not visible, the value is  `("global")` again.
- If now switching to a coordinate space `space` with is not a value in `<spatial_element>.coordinate_systems` the layer automatically becomes invisible but keeps a check of in which coordinate systems it was visible. When switching back to a coordinate system in which it is was visible, the layer automatically becomes visible again.
- To allow for the landmark registration for now a layer is allowed by the user to become visible even when not having the selected coordinate system associated with it. However, this does not affect `active_in_cs`.

In short, this PR is for tracking in which coordinate systems the layer is visible / editable and disallows visibility upon switching when the selected coordinate system is not associated with the spatial element. Until multicanvas is there the visibility can still be toggled by the user explicitly. 
This functionality requires a callback upon switching coordinate systems and upon toggling the visibility of a given layer.

melonora:
> * when the user goes to a coordinate system, all the elements that are in that coordinate system are shown and all the elements that are not in that coordinate system are hidden.
> 
> * thus, if as a user I hide a layer in a coordinate system, and I go back and forth to the same coordinate system, that layer will reappear.

I would vote against this given the number of elements that can be present in a coordinate system. With any form of multiplexed imaging this can be a lot and would be costly to have all visible.

melonora:
If the idea is also to make `active_in_cs` private, I would second that as for now that would avoid some complexity.

kevinyamauchi (commented):
@LucaMarconato , as I am going on holiday, I'm passing the review on to you. Please feel free to merge after you review/approve.

PR #102, opened 2023-06-04
Fixes/cli and points

LucaMarconato:
Fixes:
- CLI was broken. Fixed and added tests.
- Using points instead of shapes for visualizing circles, otherwise performance is bad.
- Removed gray border when showing points.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,5 +1,7 @@
+import sys
 from pathlib import Path
 
+import pytest
 from click.testing import CliRunner
 from napari_spatialdata.__main__ import cli
 from spatialdata.datasets import blobs
@@ -28,3 +30,5 @@ def test_view_path_is_dir():
         blobs().write(f)
         result = runner.invoke(cli, ["view", "data.zarr", "--headless"])
         assert result.exit_code == 0  # Command executed successfully
+        with pytest.raises(SystemExit):
+            runner.invoke(sys.exit())
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,0 +1,30 @@
+from pathlib import Path
+
+from click.testing import CliRunner
+from napari_spatialdata.__main__ import cli
+from spatialdata.datasets import blobs
+
+
+def test_view_exists():
+    runner = CliRunner()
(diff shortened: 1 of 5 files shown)


LucaMarconato:
@melonora @giovp I made this small PR to fix the CLI, unfortunately I am getting tests errors unrelated to my code. Could you please have look? It's officially my first PR in the `main` branch instead that the `spatialdata` branch 😁 so you have more experience than me with the tests. Thanks 😊

melonora:
ah this is due to qt components being connected, then one of them is being closed before the other components have finished processing and are closed. Let me check if I can reproduce locally. Solution is as stated by the error message most likely. We had a it a while back in Napari too.

giovp (approved):

giovp:
is this due to qt6? we could revert to qt5 wdyt?

melonora:
> is this due to qt6? we could revert to qt5 wdyt?

As far as I know not. The reason I think is that the way the cli is run here it is a pure cli running in a context manager. This does not properly clean up qt as there is no signal given to napari to quit so there is no clean up happening at all. You basically finish while the callbacks are still referencing and thus anything referencing objects that are not there anymore becomes stale.

tests/test_cli.py
```diff
@@ -0,0 +1,34 @@
...
+def test_view_path_is_dir():
+    runner = CliRunner()
+    with runner.isolated_filesystem():
+        f = Path("data.zarr")
+        blobs().write(f)
+        result = runner.invoke(cli, ["view", "data.zarr", "--headless"])
+        assert result.exit_code == 0  # Command executed successfully
+
+        # Close all existing viewer instances to avoid leaking of viewers between tests
+        Viewer.close_all()
```
rahulbshrestha, line 34: I added this method, offered by Napari to close all viewer instances (specifically for tests like these). This fixes the failing tests in this PR.

melonora: Nice, did not know that it would allow to close all instances like this:)

melonora (approved):

melonora:
Based on 2 approvals and tests now passing without memory leaking I will merge this one.

Issue #105, opened 2023-06-05
Extend CLI


berombau:
The napari CLI has support for loading one or more plugins. Supporting this fully is beneficial as this is what new users with napari experience will expect.

When combining CLI options, we would expect being able to open a dataset and also open the View or Scatter plugin. The following now opens two separate windows instead of one, because a new viewer instead of the current one is used.
```
napari -w napari-spatialdata View --plugin napari-spatialdata '/Users/benjaminr/Documents/GitHub/napari-spatialdata/data/cosmx/data.zarr'
```
- [ ] bug for `-w` and `--plugin`

Right now, `python -m napari-spatialdata view <path>` and `napari --plugin napari-spatialdata <path>` have the same behavior. But the first case could in the future support more advanced behavior (combining multiple datasets, filtering...).

- [x] Support CLI for multiple datasets
- [ ] Support filtering SpatialData elements

PR #106, opened 2023-06-05
Update _interactive.py

berombau:
Closes first issue in #105: bug for -w and --plugin.


rahulbshrestha:
Tested and works! thanks @berombau!

PR #107, opened 2023-06-06
Fix docs pre-commit bug

rahulbshrestha:
This PR fixes a bug in the docs that is causing a pre-commit test to fail

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -2,11 +2,12 @@
 
 The Command Line Interface can be access by running `python -m napari_spatialdata` or `napari` in a shell.
 Example:
+
 ```
 python -m napari_spatialdata view path/to/data.zarr
 ```
 
-To get extra information, run 
+To get extra information, run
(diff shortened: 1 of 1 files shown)


melonora:
it could be that the path is relative to the source path of the docs. so with `../..` you would not end up in the napari_spatialdata module. Not certain why it only started failing now though.

melonora:
ok the problem was that PyQt6 now required glibc 2.35 or higher. This can only be installed when using ubuntu 22.04 as os.

melonora:
There is an incompatibility with updating search tools at the end, this causes a warning. Therefore, I disabled the fail on warning flag

kevinyamauchi (approved):
Nice debugging! I am wondering if we could have used pyqt5 and avoided this issue. In any case, this is super easy to revert and it fixes the issue for now. I will merge and we can re-visit if necessary in the future.

Issue #110, opened 2023-06-07
pyqt6 vs pyqt5


kevinyamauchi:
Hey everyone. I just noted in #81 , we changed the Qt backend that gets installed with `napari-spatialdata[all]` to `PyQt6`. What was the motivation here?

I am concerned because many/most napari users will have Qt5 bindings installed and this may cause issues. Also, Qt6 isn't fully supported in napari.

kevinyamauchi:
cc @giovp

giovp:
totally fine for me to revert back to pyqt5

kevinyamauchi:
@LucaMarconato , you mentioned on zulip you had issues with pyqt5. What were they? Is it because you're on an M1 mac? If so, how did you install napari?

melonora:
Not by standard means. There is a possibility with homebrew, but had multiple reports of that failing for people as well. Most people when they need it as dependency open the terminal in rosetta mode as far as I know. In any case that would not be an option for us.

melonora:
https://forum.image.sc/t/issue-with-downloading-napari-on-mac-m1-chip/53961/13?mobile_view=0

kevinyamauchi:
> Not by standard means.

I am fairly certain that PyQt5 is available via conda-forge.

@LucaMarconato , have you tried `conda install -c conda-forge napari`?

I don't think we should direct users to PyQt6 when it isn't officially supported by napari yet. We also should not recommend any solution requiring homebrew.

melonora:
One thing to maybe keep in mind for future PRs to make switching later on easier is to use fully qualified names. These are most of the times readily supported both in pyqt5 and pyqt6 and make transition easier.

LucaMarconato:
CC @giovp
@melonora a comment on the code that you sent me
```bash
Step 1 - Install Anacaonda:
https://www.anaconda.com/download
Download the M1/M2 Distribution
Install the .pkg (when I did this, I was personally running Anaconda3-2023.03-1-MacOSX-arm64.pkg)

Step 2 - Setup x86 env: Source
NOTE: My env is named someRandomEnv and you should change it to whatever you would like
$ CONDA_SUBDIR=osx-64 conda create -n someRandomEnv python=3.9
$ conda activate someRandomEnv
$ conda config --env --set subdir osx-64

Step 3 - Install pyqt separately: Source
Install pyqt5
$ pip install pyqt5

Step 4 - Install Napari: Source
$ python -m pip install "napari[all]"

[...]

PR #111, opened 2023-06-09
Revert pyqt5

melonora:
As discussed in #110, we revert back to pyqt5 for now.


melonora:
@LucaMarconato When you have time, can you try with `conda install -c conda-forge pyqt`. https://anaconda.org/conda-forge/pyqt seems to support osx-arm64.

giovp:
I jusr tried and I was able to install pyqt5 via conda but not vispy.  Installing it via conda helped but still couldn't install it via pip.

LucaMarconato:
Now that this issue got closed https://github.com/napari/napari/issues/6031 I managed to install `napari` via conda (`conda-forge`), which uses `PyQt5`.

After installing `napari` via `conda`, I could then install `spatialdata` and all the other libraries (in particular `napari-spatialdata`, correctly. Here is the script that I use (not that I don't use `[all]` when installing `napari-spaitaldata` since `PyQt5` is already installed.

So, I suggest to merge this PR (which closes https://github.com/scverse/napari-spatialdata/issues/110), and update the install docs with a section on M1, explaining to first pre-install `napari` via `conda`.

```bash
#!/bin/bash                                                                                                                                      

source ~/.bash_profile
echo "-------- remove old env --------"
conda deactivate
set -e
conda remove -n ome --all -y

echo "-------- create new env --------"
mamba create -n ome python==3.10 -y
conda activate ome 

mamba install -c conda-forge napari -y
echo "-------- install spatialdata-io --------"
cd spatialdata-io
pip install -e .

echo "-------- install spatialdata-plot --------"
cd ../spatialdata-plot
pip install -e .

echo "-------- install napari-spatialdata --------"
cd ../napari-spatialdata
pip install -e . 

echo "-------- install spatialdata --------"
cd ../spatialdata
pip install -e ".[all]"

echo "-------- test --------"
cd ../spatialdata-sandbox/mibitof
python -m spatialdata peek data.zarr
python -m napari_spatialdata view data.zarr  
```

LucaMarconato (approved):

PR #114, opened 2023-06-09
Save to sdata

melonora:
This is a branch of layer_visibility_cs. It involves a refactor and rudimentary functionality to save a shapes and points layer visualizing a pre-existing spatial element. 

Saving the data makes use of the metadata field `current_cs` which indicates the current cs the element is present in. This means if currently a coordinate system in which the layer is not active is selected, the value for `current_cs` is the last coordinate system in which the layer was present. Note that currently the annotation is not saved along with it. This would require being able to keep track of which items have been adjusted in what kind of manner. I have rudimentary code ready for this but some feedback would be nice first.

There is additional code pretty much ready to go regarding saving new shapes and points layers. The PR for this I will open after initial feedback on this PR.


melonora:
Closing this in favor of new PR branched from main

Issue #80, opened 2023-05-22
information lacking in mibitof_analysis.ipynb


lopollar:
Then, we open the "Scatter" Widget by using the menu bar and go to Plugins > napari-spatialdata > Scatter. In the Scatter Widget, 

The sentence ends there.  I made at first the mistake to not have selected my shapes layer. Maybe mentioning this could help people in the future too!

PR #119, opened 2023-06-19
Update incomplete text in notebooks

rahulbshrestha:
This PR fixes the issue in https://github.com/scverse/napari-spatialdata/issues/80/.

The text has been updated from

`Then, we open the "Scatter" Widget by using the menu bar and go to Plugins > napari-spatialdata > Scatter. In the Scatter Widget,`

to 

`Select "point8_labels" in the list of layers. Then, we open the "Scatter" Widget by using the menu bar and going to Plugins > napari-spatialdata > Scatter. This loads the AnnData object associated with that layer into the "Scatter" Widget.`


LucaMarconato:
LGTM

PR #115, opened 2023-06-10
RGB images support

rahulbshrestha:
This PR fixes the issue mentioned in https://github.com/scverse/napari-spatialdata/issues/66. This is based of the code from this PR: https://github.com/scverse/napari-spatialdata/pull/93.

PR #120, opened 2023-06-22
Viewer refactor

melonora:
Currently napari-spatialdata doesn't have a viewer class on itself. This PR creates a `SpatialDataViewer` subclassing the napari `Viewer`. This allows us to have `Viewer` designated callbacks in our viewer class, preventing the widget classes from becoming too convoluted with functions that are `Viewer` specific. Most of the code related to adding spatial elements as layers has been moved to this new `SpatialDataViewer`. 

In addition when a user is adding a new layer there is no information regarding which `SpatialData` object it belongs to and neither about the coordinate systems. The new viewer class has a callback which allows for inheriting the metadata of the active layer. In napari at most one layer is active at the time, even when multiple are selected. Within the insert layer callback, the added layer is not yet the new active layer, which allows for inheriting metadata from the layer that is active just before adding a new layer.


melonora:
At the moment there is a small issue regarding the camera position not set to the center point of the data, but rather the origin.

melonora:
@kevinyamauchi @LucaMarconato I fixed the issue with the camera. Was due to me using a callback to `add_layer` which is already used in napari. Anyways it is fixed now so it is ready for review.

kevinyamauchi (commented):
Overall I like the motivation behind this PR. I think having an object that groups the napari viewer with viewer-related methods makes sense. However, I am not sure that I like subclassing the viewer. I am concerned that overriding methods on the viewer is going to have downstream consequences - especially private methods (see comment below).

Can you please explain your motivation behind using inheritance rather that creating a model class via composition?

src/napari_spatialdata/_viewer.py
```diff
@@ -0,0 +1,154 @@
...
+        layer = event.value
+        active_layer = self.layers.selection.active
+
+        if active_layer and type(layer) in {Labels, Points, Shapes} and "sdata" not in layer.metadata:
+            active_layer_metadata = active_layer.metadata
+            layer.metadata["sdata"] = active_layer_metadata["sdata"]
+            layer.metadata["_current_cs"] = active_layer_metadata["_current_cs"]
+            layer.metadata["_active_in_cs"] = {active_layer_metadata["_current_cs"]}
+            show_info(f"The spatialdata object is set to the spatialdata object of {active_layer}")
+        super()._on_add_layer(event)
```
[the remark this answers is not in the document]
melonora: The metadata inheritance should conceptually be happening as the layer gets added. Alternative callback can be `_inherit_metadata` which would not require overloading the method.

melonora:
I did check calling `napari --plugin napari-spatialdata`. This at the moment gives some problem. I will fix

src/napari_spatialdata/_viewer.py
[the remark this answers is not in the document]
kevinyamauchi: My concern is that if a user adds a layer to the viewer themselves or via another plugin, this callback will a bunch of metadata assuming it is meant to be managed by spatialdata, which may not be the case.

kevinyamauchi:
> Testing becomes more complicated I believe. Right now the make_napari_viewer fixture accept a ViewerClass. And this ViewerClass should be a callable.

I'm not sure how this makes it harder to use composition. I think composing the standard napari viewer with our plugin model will generally break fewer assumptions made by napari and the broader community about the viewer class. In the case of testing, it seems like you can make a standard napari viewer with the `make_napari_viewer` fixture and then pass that viewer to construct our model class.

> I did check calling napari --plugin napari-spatialdata. This at the moment gives some problem. I will fix. Mainly the solution of calling napari.current_viewer() does not work anymore due to methods we now have in the viewer class.

I think this is a symptom of the concerns I have about subclassing the viewer. I really think we need to consider other options before going this route.

## clarification
after re-reading the PR description, it's a bit unclear to me what the goals of the PR are. Can you please state the problems you aimed to fix with this PR?


### alternative model
Do you think having the napari viewer as a property of our spatialdata model could work instead?

```python
class SpatialDataViewer:
    def __init__(self, viewer: napari.Viewer):
        # store the viewer, may need to use napari.current_viewer() depending on how
        # things are arranged
        self.viewer = viewer

    def add_sdata_labels():

    def add_sdata_points():

    def other_useful_methods_for_managing_viewer():


class  SdataWidget(QWidget):

    def __init__(self, viewer: napari.Viewer, sdata):
        self.viewer = SpatialDataViewer(viewer)
```

melonora:
> > Testing becomes more complicated I believe. Right now the make_napari_viewer fixture accept a ViewerClass. And this ViewerClass should be a callable.
> 
> I'm not sure how this makes it harder to use composition. I think composing the standard napari viewer with our plugin model will generally break fewer assumptions made by napari and the broader community about the viewer class. In the case of testing, it seems like you can make a standard napari viewer with the `make_napari_viewer` fixture and then pass that viewer to construct our model class.
> 
> > I did check calling napari --plugin napari-spatialdata. This at the moment gives some problem. I will fix. Mainly the solution of calling napari.current_viewer() does not work anymore due to methods we now have in the viewer class.
> 
> I think this is a symptom of the concerns I have about subclassing the viewer. I really think we need to consider other options before going this route.
> ## clarification
> 
> after re-reading the PR description, it's a bit unclear to me what the goals of the PR are. Can you please state the problems you aimed to fix with this PR?
> ### alternative model
> 
> Do you think having the napari viewer as a property of our spatialdata model could work instead?
> 
> ```python
> class SpatialDataViewer:
>     def __init__(self, viewer: napari.Viewer):
>         # store the viewer, may need to use napari.current_viewer() depending on how
>         # things are arranged
>         self.viewer = viewer
> 
>     def add_sdata_labels():
> 
>     def add_sdata_points():
> 
>     def other_useful_methods_for_managing_viewer():
> 
> 
> class  SdataWidget(QWidget):
> 
>     def __init__(self, viewer: napari.Viewer, sdata):
>         self.model = SpatialDataViewer(viewer)
> ```

That works, I adjusted the code. As for the purpose of the PR:
1. Allow for metadata inheritance when creating new layers. Users ask when they do annotation whether they can save it back to the spatialdata object. On the Spatialdata branch when having multiple spatialdata object being viewed at the same time, @LucaMarconato was going for the first spatialdata object. This PR allows more control regarding that.
2. Keep the code modularized to prevent things such as we had in napari with the vispy canvas and qt being entangled.

src/napari_spatialdata/_viewer.py
```diff
@@ -0,0 +1,164 @@
...
+
+if TYPE_CHECKING:
+    from napari.utils.events.event import Event
+    from spatialdata import SpatialData
+
+
+class SpatialDataViewer:
+    def __init__(self, viewer: Viewer) -> None:
+        self.viewer = viewer
+        self.viewer.layers.events.inserted.connect(self._inherit_metadata)
```
kevinyamauchi, line 23: Thanks for the updates, @melonora ! I am still a bit concerned about this event connection. If I understand the code correctly, it the attached callback will try to add spatialdata metadata to _any_ layer added to the viewer. Is this correct?

While I can see how that would be desirable for the case where spatialdata is the only library/plugin being used with napari, I would hope that we can design our plugins such that they can be used in conjuction with other plugins and workflows. If we are automatically adding spatialdata metadata to layers added by other plugins or by the user with other intentions, that seems like it could potentially cause issues (e.g., adding spatialdata metadata to a layer where it doesn't make sense to do so).

Instead of connecting the `_inherit_metadata()` method to this event, can we call it in the `add_sdata_*()` methods? Perhaps we can also have a public version that takes a source layer and "clones" the metadata onto a target layer for the case when people want to manually add metadata? What do you think?

melonora: > Thanks for the updates, @melonora ! I am still a bit concerned about this event connection. If I understand the code correctly, it the attached callback will try to add spatialdata metadata to _any_ layer added to the viewer. Is this correct?

Yes this is correct if there is an active layer. When napari starts and no spatialdata layers have been opened yet, no spatialdata metadata can be added.

> While I can see how that would be desirable for the case where spatialdata is the only library/plugin being used with napari, I would hope that we can design our plugins such that they can be used in conjuction with other plugins and workflows. If we are automatically adding spatialdata metadata to layers added by other plugins or by the user with other intentions, that seems like it could potentially cause issues (e.g., adding spatialdata metadata to a layer where it doesn't make sense to do so).

What kind of case do you have in mind where it makes sense for the user to have a spatialdata object open without wanting an association with that object when opening a new layer? Just to have a clear idea.
> 
> Instead of connecting the `_inherit_metadata()` method to this event, can we call it in the `add_sdata_*()` methods? Perhaps we can also have a public version that takes a source layer and "clones" the metadata onto a target layer for the case when people want to manually add metadata? What do you think?

I think it would be possible by means of having source layer and target layer selected and then using a shortcut. Adding it to the `add_sdata_*()` methods would be more complicated I think. It is specifically the usecase when new layers are added to do annotation on a spatialdata object. How would you distinguish from the case you described above when a user presses the button to add a new layer which should not be associated with the spatialdata object? I think our best option would be to go for a public method triggered by a keyboard shortcut as I described if we decide to allow for layers not to be associated with a spatialdata object.

Another question is also whether we should allow for 2 spatialdata objects to be associated with one layer.

melonora:
> @melonora , sorry, but merging #115 caused a merge conflict

No worries, I will pick this and the notebooks up today.

src/napari_spatialdata/_viewer.py
kevinyamauchi: > What kind of case do you have in mind where it makes sense for the user to have a spatialdata object open without wanting an association with that object when opening a new layer? Just to have a clear idea.

One example is that some plugins create temporary layers for annotations or displaying visuals. I don't think these should be associated with the spatialdata object. In general, I don't think we should start by automatically adding metadata to all layers. We can add it in the future as necessary, but I don't think we should be careful about adding "magic" at this early stage.

> I think our best option would be to go for a public method triggered by a keyboard shortcut as I described if we decide to allow for layers not to be associated with a spatialdata object.

I agree this is the best approach. We can have a public helper function that adds a layer for annotation (i.e., adds the layer and assigns the spatialdata metadata) and potentially some UI buttons in the future.

melonora:
@kevinyamauchi @LucaMarconato Ok so couple of comments. I did create a keybind which calls a private method, which in turn calls a public method. The public method can also directly be called from the console. It takes as argument `layers`. In case of being called by private method this is a list of selected layers and in case of directly calling the public method this is a list of layers defined by the user. Two questions:
1. Labels / Points and Shapes also hold specific anndata metadata, some of which was under discussion to be removed. Should I still implement that this metadata gets created?
2. At the moment for some reason when calling the inherit function twice with the same selected layers, `sdata_count` becomes 2. I don't know exactly why this is the case and how to prevent it. It seems that assignment creates a copy rather than an extra reference to the same memory address. Any ideas?

kevinyamauchi:
> Labels / Points and Shapes also hold specific anndata metadata, some of which was under discussion to be removed. Should I still implement that this metadata gets created?

I am not sure about the exact metadata, but I would keep it if we haven't officially removed it yet.

src/napari_spatialdata/_viewer.py
```diff
@@ -0,0 +1,189 @@
...
+        Paramters
+        ---------
+        layers: list[Layer]
+            A list of napari layers of which only 1 should have a spatialdata object from which the other layers inherit
+            metadata.
+        """
+        # Layer.metadata.get would yield a default value which is not what we want.
+        sdatas = [layer.metadata["sdata"] for layer in layers if "sdata" in layer.metadata]
+        # len(sdatas) as opposed to 1 to ensure sdatas is not empty
+        sdata_count = sum(sdatas[0] is sdata for sdata in sdatas[1:]) if len(sdatas) != 1 else len(sdatas)
```
kevinyamauchi, line 45: > At the moment for some reason when calling the inherit function twice with the same selected layers, sdata_count becomes 2. I don't know exactly why this is the case and how to prevent it. It seems that assignment creates a copy rather than an extra reference to the same memory address. Any ideas?

I am sorry, but I am not sure I understand your test case. Is this what you did?

1. select 2 layers. one layer has SpatialData metadata, the other does not
2. you call the `inherit_metadata()` function 
3. With those same two layers selected, you call the `inherit_metadata()` function again
4. `sdata_count` is 2

If so, it seems like there are a couple of things wrong:

- line 47 to me suggests that sdata_count was meant to count the number of unique SpatialData objets. On the other hand, this count looks like it's counting the number of times the first spatial data object is the same as the other spatial data objects.
-  it doesn't seem to me like the count here should be 2 in the case above. it seems like it should be 1 since there would be 2 sdata objects and iterating starts at 1 (sdatas[1:]), the max value would be 1. I would check to make sure you're passing the layers you meant to into the `inherit_metadata()` function.

melonora: I assumed that one might quickly select all layers in which case the behaviour is as follows assuming this test case:
1. Open a napari spatialdata with a SpatialData object.
2. Open 2 elements as layers, these are backed by the same Spatialdata object.
3. Open a new annotation layer, for example Shapes.
4. Select all layers and press `shift+L`.

Now the code does the following:
1. Get all sdata objects from the metadata in layers as list.
2. Check whether they all point to the same memory address -> sdata_count != 1 if no element in `sdatas` or more than 1 spatialdata memory address.
3. If check passes, get first layer with `sdata` in metadata.
4. Let annotation layers with no `sdata` in metadata inherit from the layer in step 3.

This still works fine up to this point. The shapes layer properly inherits from one of the layers that has `sdata`. Now when pressing `shift-L` again when the same layers are still selected we get the behaviour that `sdata_count` now is 2, which in my opinion shouldn't be the case. It seems that when inheriting the metadata a copy of `sdata` is created and thus there are now 2 memory addresses having a spatialdata object even though it should be one address.

The case for checking whether there are more than 2 addresses is because if it is the case than we don't necessarily know which spatialdata object to associate with the new annotation layer.

melonora:
Regarding metadata @LucaMarconato, could you please pitch in as it touches also the issue of plotting shapes when there is no table. The question is basically whether there should always be a table present or not.

src/napari_spatialdata/_viewer.py
melonora: > Hey @melonora , I am sorry, but I am having trouble reconciling your comment, the code, and your question above. Can you please write a test using the API (i.e., not with the keyboard shortcut) that covers this case so I can see where the current behavior is not meeting your expectations?

I will

melonora: The main thing is that if we add sdata metadata to a layer from another layer, we should not have it that it seemingly points to 2 different spatialdata objects in memory, which it now seemingly does. Funnily enough when just opening multiple spatial elements backed by the same spatialdata object as layers in the viewer as you can see from the test then the count is 1.

melonora: Also relocating this function so we directly use widget._sdata to see if it solves the problem is not an option as I think this likely will go away with a follow up PR regarding supporting multiple spatialdata objects.

melonora: Just checked with id, all objects do have the same id so I will adjust the code to use id

melonora:
@kevinyamauchi I briefly discussed with @LucaMarconato, since there are no associated geometries yet it could be better with regards to anndata here. This should become metadata when actually saving to SpatialData.

src/napari_spatialdata/_viewer.py
```diff
@@ -0,0 +1,188 @@
...
+        Paramters
+        ---------
+        layers: list[Layer]
+            A list of napari layers of which only 1 should have a spatialdata object from which the other layers inherit
+            metadata.
+        """
+        # Layer.metadata.get would yield a default value which is not what we want.
+        sdatas = [layer.metadata["sdata"] for layer in layers if "sdata" in layer.metadata]
+
+        sdata_ids = {id(sdata) for sdata in sdatas}
```
kevinyamauchi, line 45: I don't think we should use `id()` (see post below). If I understood correctly, you switched to `id()` because there were some issues with the counting of unique sdata objects. If I recall correctly, your previous sdata count was potentially off because you were doing something like `sum(sdata_a is sdata_b)`, which would count the number of times `sdata_a` is repeated, not the number of unique sdatas there are. 

https://stackoverflow.com/questions/52268343/id-vs-is-operator-is-it-safe-to-compare-ids-does-the-same-id-mean-the

kevinyamauchi (changes requested):
Thanks for the updates, @melonora ! As mentioned below, I don't think we should use `id()`. Can you please change to `is`?

> The main thing is that if we add sdata metadata to a layer from another layer, we should not have it that it seemingly points to 2 different spatialdata objects in memory, which it now seemingly does. Funnily enough when just opening multiple spatial elements backed by the same spatialdata object as layers in the viewer as you can see from the test then the count is 1.

What makes you think that adding the sdata object to another layer copies it (i.e., creates a new object with a different memory address)? I don't think this is the case if all objects have the same `id()`.

> I briefly discussed with @LucaMarconato, since there are no associated geometries yet it could be better with regards to anndata here. This should become metadata when actually saving to SpatialData.

I am sorry, I do not know what you mean. Can you please clarify? I think I am missing some context.

melonora:
> Thanks for the updates, @melonora ! I think this is close. As mentioned below, I don't think we should use `id()`. Can you please change to `is`?
> 
> > The main thing is that if we add sdata metadata to a layer from another layer, we should not have it that it seemingly points to 2 different spatialdata objects in memory, which it now seemingly does. Funnily enough when just opening multiple spatial elements backed by the same spatialdata object as layers in the viewer as you can see from the test then the count is 1.
> 
> What did you observe that makes you think that adding the sdata object to another layer copies it (i.e., creates a new object with a different memory address)? I don't think this is the case if all objects have the same `id()`.

Yeah it was a silly mistake of me using `sum`. I don't know what I was thinking at the moment. I switched back to `is` and combined it with `all` to ensure all `SpatialData` objects are the same if more than 1 layer has a `SpatialData` reference in the metadata.
> 
> > I briefly discussed with @LucaMarconato, since there are no associated geometries yet it could be better with regards to anndata here. This should become metadata when actually saving to SpatialData.
> 
> I am sorry, I do not know what you mean. Can you please clarify? I think I am missing some context.

So Points, Shapes and Label layers have `AnnData` metadata associated, for example:
```
"adata": sdata.table[sdata.table.obs[sdata.table.uns["spatialdata_attrs"]["region_key"]] == key],
                "shapes_key": sdata.table.uns["spatialdata_attrs"]["region_key"],
```
The question was whether this should be newly created as well when a user decides to inherit spatialdata metadata. However, at the point of time the user might not have created any annotations yet. So the correct time for this to happen would be when saving layers to the SpatialData object. Or at least that is what we think.

tests/test_cli.py
```diff
@@ -32,3 +33,4 @@ def test_view_path_is_dir():
 
         # Close all existing viewer instances to avoid leaking of viewers between tests
         Viewer.close_all()
+        shutil.rmtree(f)
```
LucaMarconato, line 36: Do we need this? isn't this deleted by default? In general we try to avoid to use `shutils` for safety reasons

LucaMarconato:
> Labels / Points and Shapes also hold specific anndata metadata, some of which was under discussion to be removed. Should I still implement that this metadata gets created?

As Kevin said, I would keep the `AnnData` metadata for the moment.

src/napari_spatialdata/_viewer.py
```diff
@@ -0,0 +1,187 @@
...
+class SpatialDataViewer:
+    def __init__(self, viewer: Viewer) -> None:
+        self.viewer = viewer
+        self.viewer.bind_key("Shift-L", self._inherit_metadata)
+
+    def _inherit_metadata(self, viewer: Viewer) -> None:
+        layers = list(viewer.layers.selection)
+        self.inherit_metadata(layers)
+
+    def inherit_metadata(self, layers: list[Layer]) -> None:
```
LucaMarconato, line 29: I think this approach is a clean and safe one, so I would merge the PR now. But I wonder if we can avoid to have the user having to know that this shortcut exists, to make the user experience more intuitive. We could explore this in a follow up PR so that we unlock this one.

One idea that we could consider for instance is to take advantage that the user will have to press `Shift+E` to save the annotations to disk. So maybe we could go back to place the metadata in every layer that is added, but ignore the metadata in those layers. 

So for instance, even if an external plugin is adding layers which then will be populated with some metadata, in particualar `_active_in_cs`, practically no behavior will be induced by the presence or absence of those metadata. But when the user presses `Shift+E` to save a layer, then those metadata will become "active", and we will consider them.

Wdyt?

LucaMarconato:
Thanks @melonora for the PR, I think it makes the structure of the plugin much cleaner, and thanks @kevinyamauchi for the extensive discussion and valid points raised. I gave also my review, and for me the PR is ready to merge after the comment on `shutil` is addresed and with an eventual follow PR on the shortcut usage.

Apologies if I gave the review late, I was catching up with many notifications and took a while to arrive to this PR. I talked with Wouter about some points via Zulip, but managed to complete the review only today.

src/napari_spatialdata/_viewer.py
melonora: I am unsure whether I understand you correctly, but I think this would be quite challenging from based on my current understanding of your proposal. Differentiating between layers that are already backed by a `SpatialData` storage and those that are not in that case would require an additional check before applying a certain behavior that is currently there. It also forces the user to practically save first in order to make particular behaviour available.

That being said, in the long run I think we can avoid this completely with true multicanvas view where we have the `SpatialData` layers on one canvas and anything else on the other.

src/napari_spatialdata/_viewer.py
melonora: resolved with [e4761cf](https://github.com/scverse/napari-spatialdata/pull/120/commits/e4761cf5263b2b6191c0cbc668b0447a8ec1ba8c)

tests/test_cli.py
melonora: This was a remnant from having 2 distinct tests both writing a `SpatialData` object to the same location. Given the issue with overwrite this caused the tests to fail. However since then one of these tests is not there anymore so it is currently not necessary. Resolved with ([7ebcce5](https://github.com/scverse/napari-spatialdata/pull/120/commits/7ebcce54d0ff378cad824088c8cebbe18af645dd)

melonora:
> > Labels / Points and Shapes also hold specific anndata metadata, some of which was under discussion to be removed. Should I still implement that this metadata gets created?
> 
> As Kevin said, I would keep the `AnnData` metadata for the moment.

I mean not getting rid of this part of the metadata getting created. Rather, should this metadata only be created upon saving a new layer to the zarr store of a `SpatialData` object or upon linking the new layer to a `SpatialData` object.  I am not certain whether the latter would be favorable if in general `SpatialElement` not necessarily having a table is supported.

src/napari_spatialdata/_viewer.py
LucaMarconato: Ok, so then let's keep the currently implemented behavior and see later how to remodel it with multicanvas.

LucaMarconato:
> I mean not getting rid of this part of the metadata getting created. Rather, should this metadata only be created upon saving a new layer to the zarr store of a SpatialData object or upon linking the new layer to a SpatialData object. I am not certain whether the latter would be favorable if in general SpatialElement not necessarily having a table is supported.

I would say upon linking, to keep the behavior consistent (everything happens during linking and not saving). In case in which no table is present I would just put an empty table or a `None` in the metadata.

melonora:
> > I mean not getting rid of this part of the metadata getting created. Rather, should this metadata only be created upon saving a new layer to the zarr store of a SpatialData object or upon linking the new layer to a SpatialData object. I am not certain whether the latter would be favorable if in general SpatialElement not necessarily having a table is supported.
> 
> I would say upon linking, to keep the behavior consistent (everything happens during linking and not saving). In case in which no table is present I would just put an empty table or a `None` in the metadata.

I added additional metadata with as value `None`. This would be updated in a later PR upon saving.

melonora:
Instead of using `labels_key` and `shapes_key` we use `region_key` to have a more uniform API. Test was initially failing because of still using `label_key`. Should be fixed now.

src/napari_spatialdata/_viewer.py
kevinyamauchi: I agree it would be nice to do this automatically. However, I think it's difficult to do so at the moment, so I like the approach of having a manual approach backed by a public method that can be used whenever we figure out how to do it automatically.

kevinyamauchi (approved):
LGTM! Thanks for the work and for your patience, @melonora !

@LucaMarconato , please take a final look and merge when you're ready.

kevinyamauchi:
@melonora said that @LucaMarconato approved the PR in person, so I will merge now. Thanks everyone!

Issue #126, opened 2023-06-29
Slow load time layers


melonora:
Possibly related to #125, loading layers seems to take longer than expected. We should check bottlenecks.

LucaMarconato:
Please see my comment in: https://github.com/scverse/napari-spatialdata/issues/125, not sure that's the cause, but definitely it's a bottleneck.

Issue #125, opened 2023-06-29
Issue: slow update time element widget.


melonora:
When layers are being displayed and then switching coordinate systems, it takes a long time to update the element widget.

LucaMarconato:
by contrast, the speed is very good in the `spatialdata` branch

LucaMarconato:
I think it's due to this https://github.com/scverse/napari-spatialdata/blob/e1d308d16a8dcbb7b23de4473de08c1d698b7212/src/napari_spatialdata/_interactive.py#L239

We should pass a list of images to `add_image()` to make use of the multiscale representation.

PR #129, opened 2023-07-02
Fixes performance with multiscale, fixes docs

LucaMarconato:
* Fixes performance issue with multiscale images. Performance tested on the Xenium + Visium dataset (the one where the performance problem got reported)
	* closes https://github.com/scverse/napari-spatialdata/issues/126
	* closes https://github.com/scverse/napari-spatialdata/issues/125
* Refactor the name and the docstring of `_transform_to_rgb()`, now called `_adjust_channels_order()`, as the name was misleading. The function is not transforming to RGB, but it's swapping the dims and telling if the image is RGB or not. Also the swap of the dims is done all the time a `c` dimension is present and not just when an RGB image is passed, as the old docstring suggested.

### DIFF
--- a/src/napari_spatialdata/_interactive.py
+++ b/src/napari_spatialdata/_interactive.py
@@ -7,12 +7,11 @@ import numpy as np
 import shapely
 from anndata import AnnData
 from loguru import logger
-from multiscale_spatial_image import MultiscaleSpatialImage
 from napari.viewer import Viewer
 from qtpy.QtWidgets import QLabel, QListWidget, QListWidgetItem, QVBoxLayout, QWidget
 from spatialdata import SpatialData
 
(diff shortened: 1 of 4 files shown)


melonora (approved):

melonora (changes requested):
```suggestion
assert (widget._viewer.layers[0].data == widget._viewer.layers[1].data._data[0]).all()
```
When `MultiscaleSpatialImage` this ` widget._viewer.layers[1].data` is not the proper way to access the first level of the data. The suggestion above fixes the test in `test_spatialdata.py` and then we should be good for merge.

LucaMarconato:
Done!

melonora (approved):

PR #133, opened 2023-07-13
added m1/m2 installation instructions

LucaMarconato:


melonora (approved):
I am ok with the text, but I don't have a Mac to test myself. Perhaps @giovp is able to?

PR #134, opened 2023-07-13
tiny fix cli

LucaMarconato:
Old branch mostly already merged. Makes a small fix in the docs and in the CLI script.


melonora:
Just for the record are we sticking with this style of docstrings?

melonora (changes requested):
On second thought, I see in `pyproject.toml` that we follow the `numpy` docstyle which is not followed in the `view` function.

LucaMarconato:
Thanks for the review, I changed the docstring style.

> I will approve for now, but I should have the PR for multiple spatialdata objects ready today, so maybe not worth merging.

the actual changes are really small, so merging will not interfere with new PRs.

PR #136, opened 2023-07-14
Support multiple SpatialData objects

melonora:
This PR allows for viewing multiple `SpatialData` objects with napari-spatialdata. This, by passing a list of `SpatialData` objects to `Interactive`. This list is converted to an `EventedList`. In order to properly support viewing elements even when `SpatialElement` names are shared between `SpatialData` objects a few changes have been made.

1. The' ElementWidget' checks whether there are duplicate elements which are stored in the list `element_names`, If `element_name` is in this `list` then a suffix `_{index}` is added to `element_name`. The entry of `element_name` in `_elements` is changed from a `str` corresponding to the element type, to a dictionary with key-values 
         -  `element_type` corresponding to the original value of `_elements, `
         - `sdata_index` corresponding to the index of the `SpatialData` object in the `EventedList`
         - `original_name` corresponding to the original name.
 This behavior also ensures that instead of napari creating an index for the copy which is not informative, a suffix will be pointing to the `SpatialData` object. This also means in case of more than 1 duplicate `SpatialElement` name, both layers belonging to the same `SpatialData` will have the same suffix. The way it is implemented can also handle duplicate `element_name` with different associated coordinate systems.
2. A new `_get_sdata` function is used to obtain the `SpatialData` object from the `EventedList` and to indicate whether multiple `SpatialElements` with the same name exist.
3. `SdataWidget` and `SpatialDataViewer` make use of the function at 2 to create the layers appropriately. In the viewer, the `name` argument now corresponds to the name with a potential suffix, and `name` in metadata is the name in the `SpatialData` object.


src/napari_spatialdata/utils/_utils.py
```diff
@@ -279,3 +281,29 @@ def _adjust_channels_order(element: SpatialImage | MultiscaleSpatialImage) -> tu
...
+    ----------
+    sdata: EventedList
+        EventedList containing the SpatialData objects currently associated with the viewer.
+    elements
+    key: str
+        The name of the item in the element widget.
+
+    Returns
+    -------
+    tuple[int, str]
```
LucaMarconato, line 300: the return is of a different type

src/napari_spatialdata/utils/_utils.py
```diff
@@ -279,3 +281,29 @@ def _adjust_channels_order(element: SpatialImage | MultiscaleSpatialImage) -> tu
...
+
+def _get_sdata_key(sdata: EventedList, elements: dict[str, dict[str, str | int]], key: str) -> tuple[SpatialData, bool]:
+    """
+    Get the index of SpatialData object and key of SpatialElement.
+
+    Parameters
+    ----------
+    sdata: EventedList
+        EventedList containing the SpatialData objects currently associated with the viewer.
+    elements
```
LucaMarconato, line 294: please document this, could you also make an example of what is being passed?

src/napari_spatialdata/utils/_utils.py
```diff
@@ -279,3 +281,29 @@ def _adjust_channels_order(element: SpatialImage | MultiscaleSpatialImage) -> tu
         new_raster = list_of_xdata
 
     return new_raster, rgb
+
+
+def _get_sdata_key(sdata: EventedList, elements: dict[str, dict[str, str | int]], key: str) -> tuple[SpatialData, bool]:
```
LucaMarconato, line 286: general question, if the user renames some layers or duplicates some layers, this function will still operate correctly right?

src/napari_spatialdata/_sdata_widgets.py
```diff
@@ -1,53 +1,74 @@
...
     def _onClickChange(self, selected_coordinate_system: QListWidgetItem | int | Iterable[str]) -> None:
         self.clear()
 
         elements = {}
-        for element_type, element_name, _ in self._sdata.filter_by_coordinate_system(
-            selected_coordinate_system
-        )._gen_elements():
-            elements[element_name] = element_type
+        element_names = [element_name for sdata in self._sdata for _, element_name, _ in sdata._gen_elements()]
+        element_names = [element for element, count in Counter(element_names).items() if count > 1]
```
LucaMarconato, line 29: please rename to duplicate_element_names

src/napari_spatialdata/_sdata_widgets.py
```diff
@@ -97,46 +118,55 @@ def _update_layers_visibility(self) -> None:
...
     def _add_shapes(self, key: str) -> None:
-        if type(self._sdata.shapes[key].iloc[0][0]) == shapely.geometry.point.Point:
-            self._add_circles(key)
-        elif (type(self._sdata.shapes[key].iloc[0][0]) == shapely.geometry.polygon.Polygon) or (
-            type(self._sdata.shapes[key].iloc[0][0]) == shapely.geometry.multipolygon.MultiPolygon
+        sdata, multi = _get_sdata_key(self._sdata, self.elements_widget._elements, key)
+
+        original_name = key
+        if multi:
+            original_name = original_name[: original_name.rfind("_")]
```
LucaMarconato, line 146: I'd add some validations here, like that what is at the right side of _ is a number. This to recognize cases like this:

1. The user adds two spatialdata objects
2. The user removes some layers, in particular "my_layer_0"
3. The user manually renames "my_layer_1" to "my_layer".
4. Here original name would lead to "my", leading to a wrong behavior

src/napari_spatialdata/_viewer.py
```diff
@@ -65,21 +67,34 @@ def inherit_metadata(self, layers: list[Layer]) -> None:
 
         show_info(f"Layer(s) without associated SpatialData object inherited SpatialData metadata of {ref_layer}")
 
-    def add_sdata_image(self, sdata: SpatialData, selected_cs: str, key: str) -> None:
-        affine = _get_transform(sdata.images[key], selected_cs)
-        rgb_image, rgb = _adjust_channels_order(element=sdata.images[key])
+    def add_sdata_image(self, sdata: SpatialData, selected_cs: str, key: str, multi: bool) -> None:
+        original_name = key
+        if multi:
+            original_name = original_name[: original_name.rfind("_")]
```
LucaMarconato, line 73: same coment here regarding the validation. I think this code should be put in a function that does such validation.

src/napari_spatialdata/utils/_utils.py
LucaMarconato: Actually better to leave the example out since it's a private function and one can read the code anyway.

src/napari_spatialdata/utils/_utils.py
LucaMarconato: No I think it would not, see one of my other comments (the one on `rfind()`)

LucaMarconato (commented):
Added more comments.

src/napari_spatialdata/utils/_utils.py
melonora: This would still work correctly as this function is only called when adding a layer when loading spatialdata objects by calling Interactive.
The layers are not loaded yet at the stage when this function is called. A user can change the layer name, but as far as I understand we wouldn't necessarily save
that back to SpatialData yet, so the element name is still valid. Just the user would lose the hint
to which SpatialData object it belongs.

src/napari_spatialdata/_sdata_widgets.py
melonora: This function is only called when adding the layer. Given that the `element_name` is created in `SdataWidget` this should hold true still. A user
can only change the layer name at the moment, but this does not change the element name. Only when saving a 
renamed layer back do we need to perform extra checks.

src/napari_spatialdata/_viewer.py
melonora: See above

src/napari_spatialdata/utils/_utils.py
melonora: Resolved with [020c28a](https://github.com/scverse/napari-spatialdata/pull/136/commits/020c28a3da3a91463c4a10ca42d3fc6d4307cbdb)

src/napari_spatialdata/utils/_utils.py
melonora: Resolved with [020c28a](https://github.com/scverse/napari-spatialdata/pull/136/commits/020c28a3da3a91463c4a10ca42d3fc6d4307cbdb)

src/napari_spatialdata/utils/_utils.py
melonora: In other words this function only deals with the elements in the `ElementWidget` and not with layers.

src/napari_spatialdata/_sdata_widgets.py
melonora: Resolved with [d2d66f3](https://github.com/scverse/napari-spatialdata/pull/136/commits/d2d66f32507a57d1adcecd1c93022e02b49ebc4d)

src/napari_spatialdata/utils/_utils.py
```diff
@@ -291,13 +291,15 @@ def _get_sdata_key(sdata: EventedList, elements: dict[str, dict[str, str | int]]
...
+        Dictionary from elements widget containing the keyname as keys and a dictionary with the type of element, index
+        of the SpatialData object and the original name in the SpatialData object.
     key: str
         The name of the item in the element widget.
 
     Returns
     -------
-    tuple[int, str]
+    tuple[SpatialData, bool]
         The index of the SpatialData object in the EventedList and the name of the element in the
```
LucaMarconato, line 303: Thanks for the commit. I noticed also this string should be corrected.

src/napari_spatialdata/utils/_utils.py
LucaMarconato: I'm not sure I understood this 100%, but I explain the use case I have in mind. Say that we add a shapes layer (polygons) to napari. If we want to plot annotations on top of it we need to know two things: 
1) the `GeoDataFrame` object (because we need the index and maybe also additional columns)
2) the `table` in the associated `SpatialData` object.
This information is not needed just when creating the layer, but any time the user wants to visualize some numerical/categorical annotations.

Saving the index of the `SpatialData` object and the `name` of the element is fragile upon data renaming and removing/adding new `SpatialData` object. I would therefore add as metadata the `GeoDataFrame` (or other type of element) and the original `SpatialData` object.

This implies that we don't need to save `element_type` (since it can be deduced on the fly, and we don't need `sdata_index`. 

Remaining `original_name`, we can store it, or we can maybe deduce it on the fly by removing the last `_[0-9]+$` from the name. Not sure which of these two approaches is best.

melonora: So currently the layers do have the `SpatialData` object associated with it in the metadata. Changing the layer name has no effect here. Upon saving the `SpatialData` object in the metadata will be used. This would have 1 path associated with it, so even if saving multiple layers sequentially, it does not matter that the metadata for sdata is not updated yet. Or we need to update corresponding `sdata` metadata after each layer is saved, but I think because of using references to one object this is not the case.

The only reason for using the index of `sdata` is to provide the user a clue to which `SpatialData` object in the `EventedList` a layer belongs. The reason for this being that we don't have an identifier that is easily visible for the user when having multiple layers open that would otherwise have a naming collision. Napari's behaviour for naming collisions could lead to confusion for the user.

In this PR we don't give the option yet for adding or removing a `SpatialData` object using a method in `Interactive`. When we do in the follow up PR I think the index of the layer names can be easily determined should there be a need (only in case there are duplicated element names). This can just be done by looking what the index is of `sdata` in metadata in the `EventedList` and then performing a layer rename.

melonora: Also, element type is already used on main just to deduce what kind of layer should be created. The functionality here did not change other than that now it is a key with a value in a dictionary while before it was directly a value in `_elements`.

giovp: this is quite complicated to follow in terms of user behaviour and use cases.

> Saving the index of the SpatialData object and the name of the element is fragile upon data renaming and removing/adding new SpatialData object. 

Why would we want the user to rename a napari layer? I would actually even consider if possible to disallow that at the GUI level (prob difficult). it's also in scope with this being an interactive viewer that at most "add" elements but not rename name (especially not in GUI).

> I would therefore add as metadata the GeoDataFrame (or other type of element) and the original SpatialData object.

This doesn't feel right and has the potential to create a lot of chaos imho

src/napari_spatialdata/_sdata_widgets.py
giovp: > A user can only change the layer name at the moment, but this does not change the element name. 

why do we want the user to rename a layer? I would really discourage that.

src/napari_spatialdata/_sdata_widgets.py
```diff
@@ -1,53 +1,74 @@
...
+                        "sdata_index": index,
+                        "original_name": element_name,
+                    }
+                else:
+                    new_name = element_name + f"_{index}"
+                    elements[new_name] = {
+                        "element_type": element_type,
+                        "sdata_index": index,
+                        "original_name": element_name,
+                    }
```
giovp, line 48:
```suggestion
        for index, sdata in enumerate(self._sdata):
            for element_type, element_name, _ in sdata.filter_by_coordinate_system(
                selected_coordinate_system
            )._gen_elements():
            	elements_metadata = {
                        "element_type": element_type,
                        "sdata_index": index,
                        "original_name": element_name,
                    }
                name = element_name if element_name not in duplicate_element_names else element_name + f"_{index}"
                elements[name] = elements_metadata
```

giovp (changes requested):
very important addition in terms of functionality, but agree with Luca that there should be some validation (or disallow to rename layer) to improve robustness

src/napari_spatialdata/utils/_utils.py
melonora: To answer your question, I don't think we would like the user to do this, except when users add a new layer for ROI annotation. For example, someone opens their image data to annotate artefacts using a shapes layer that they add. By default this layer gets the name shapes. A user would probably like to rename this, then use the linking functionality that I implemented to link the layer to a particular `SpatialData` object and then save it.
The case that I do catch here is when users add multiple spatialdata objects in which some `SpatialElements` share names. Because the original name in the `SpatialData` object is part of `layer.metadata` this is safe to do so. It doesn't have any other effect as far as I can tell then to provide the user with a clue in case of a name collision, which `SpatialData` object the layer belongs to.

One thing to note is that `sdata` is already in `layer.metadata` so adding the GeoPandas df would be overkill.

In any case I do not think the functionality would be broken the way I implemented it currently. I think we are talking about edge cases where we are not even certain whether users would run into this issue. I would vote in this case for first seeing whether users are actually going to run into this issue.

src/napari_spatialdata/_sdata_widgets.py
```diff
@@ -1,53 +1,74 @@
...
 
 class CoordinateSystemWidget(QListWidget):
-    def __init__(self, sdata: SpatialData):
+    def __init__(self, sdata: EventedList):
         super().__init__()
 
         self._sdata = sdata
 
-        self.addItems(self._sdata.coordinate_systems)
+        coordinate_systems = {cs for sdata in self._sdata for cs in sdata.coordinate_systems}
```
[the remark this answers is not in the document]
melonora: It is the case that multiple spatialdata objects can have the same coordinate system. If a user is displaying 2 `SpatialData` objects and in it they have the same coordinate system isn't it a usecase for those 2 `SpatialData` objects to be aligned in that coordinate system?

I think I do understand your worry when it comes to coordinate systems where this isn't the case, for example the current `global`, but I am not certain whether there is a better way if we provide the functionality to view more than 1 `SpatialData` object at once. If we want to separate this completely I think it requires a widget displaying each `SpatialData` object.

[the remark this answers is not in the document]
giovp: good point yes I agree, maybe there could be a specific behaviour for when the intrinsic coordinate system is referenced

melonora:
I have added additional behaviour in order to validate `layer.name` upon a `layer.name` change. The following has changed:

1. Add callbacks for when layers are added / removed. Inserted layers have a callback being connected for when their name is changed.
2. Add self.layer_names which is updated when layers are inserted / removed and when a `layer.name` is changed. This is added to be able to determine what the layer name was before a change.
3. A name validation callback:
    - if `sdata` is in the metadata and the name is already in `sdata` then the name change is reverted.
    - if `sdata` is in metadata and `layer.name` is changed to a name of an already open layer then the name change is reverted.
    - if `sdata` and the new `layer.name` is in any of the `SpatialData` objects, name is changed to `layer.name + _sdata.index`
    - If no `sdata` in metadata, but a layer with same name already exists or the name is found in any of the `SpatialData` objects, the name is reverted in order to prevent potential naming conflicts. This should prevent issues when linking layers to a `Spatialdata` object.
    
    Currently, these behaviours have been tested manually. I am looking to add tests, but currently the event blocker being used does not seem to work in the tests so I am trying to figure that out. However, the code can already be reviewed.

src/napari_spatialdata/utils/_utils.py
LucaMarconato: Ok thanks for the info, I'd actually go for your approach.

src/napari_spatialdata/_sdata_widgets.py
LucaMarconato: > why do we want the user to rename a layer? I would really discourage that.

As Wouter said this could happen due to the napari design, but actually reading the previous Wouter's comment, it should not be a problem since the new name would not be used internally. So I think that this issue can be resolved.

src/napari_spatialdata/_viewer.py
LucaMarconato: 👍

src/napari_spatialdata/_sdata_widgets.py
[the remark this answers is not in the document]
LucaMarconato: I confirm that the current approach is correct: it is assumed that if the name is the same then the objects are aligned. The `global` space will disappear with the transformation refactoring, so this code is future proof.

LucaMarconato:
Thanks for the additional validations @melonora, I will review the changes and add comments in the next days (but in case anyone reviews before me and approves, please proceed with merging).

src/napari_spatialdata/utils/_utils.py
LucaMarconato: Minor thing: this has not been addressed yet.

LucaMarconato (commented):
I re-reviewed the updated code, thanks for the work. The code looks good to me I would merge it, but then after trying it on the Xenium + Visium dataset it fails in displaying the data.

There was a small bug in the CLI but fixing it didn't solve the problem. I still think it should be something small and not of the logic you implemented, which seems good to me.

I will debug it and push the fixes. For the records, this is the command that fails (the data shown here is produced by running [this notebook](https://github.com/scverse/spatialdata-notebooks/blob/main/notebooks/paper_reproducibility/00_xenium_and_visium.ipynb) on Visium + Xenium integration).

```bash
python -m napari_spatialdata view xenium_rep1_io/data_aligned.zarr xenium_rep2_io/data_aligned.zarr/ visium_associated_xenium_io/data_aligned.zarr
```

LucaMarconato:
Fixed the docstring and the CLI bug, they were indeed tiny. Unfortunately there is a more serious problem related to the addition of the suffixes _0, _1, arising when adding a shapes element:

image

LucaMarconato:
Added the fixes, and thanks for yours on the table. There were also some bugs (older, not related to this PR) on the display of points and circles. I fixed them.

I could test it and it works great. 
- [ ] There is only one bug that I found that is still open (checkbox to keep track things to do in this PR), here is how to reproduce it:
1. open `transcripts_1`
2. open`transcripts_2`
3. plot `qv` on top of `trancript_2`
4. click on `trancript_1`
5. try to plot `qv` also here

Then I get the following error:
image

Same things for shapes, so any small test dataset with a duplicate element that has annotations either in a table either in a column of the Dask DataFrame / GeoDataFrame should be able to reproduce the bug. Could you please have a look?

LucaMarconato:
Works great! Thanks for the fix. Good to merge now.

A minor things, I get this warning when I add or select elements),do you also get it?
image
Do you have any quick fix? If so it would be great if it could be quickly implemented, otherwise it can also be tracked for a follow up PR.

melonora:
> Works great! Thanks for the fix. Good to merge now.
> 
> A minor things, I get this warning when I add or select elements),do you also get it? image Do you have any quick fix? If so it would be great if it could be quickly implemented, otherwise it can also be tracked for a follow up PR.

Yes, there is one thing I am testing is whether we can switch to using `active` instead of `_current`. It will become deprecated, see #137. The main thing is that when having multiple selected items, you can still have a `_current` item, but not `active`. The `active` item is always `_current` though.

giovp (approved):

giovp:
it looks great @melonora ! LGTM!

Issue #123, opened 2023-06-28
Make a function to open napari with every layer shown


LucaMarconato:
We should make a function that opens napari-spatialdata with every layer shown (and ideally showing the first gene from var and first column from obs). This function would be used during automatic notebook testing, since if we open napari it doesn't show any element by default, so an eventual bug would less likely be caught.

melonora:
@LucaMarconato you think this can be closed now?

PR #140, opened 2023-07-21
Improved ergonomics of `Interactive`

LucaMarconato:
Small PR.
- [x] added the argument `headless: bool = False` to `Interactive.__init__()`. When `False`, the initializer calls `self.run()`. This is a small ergonomics improvement for the user since `interactive.run()` will be always called, and if for some reasons is not needed (like tests with napari being healess), this is still possible thanks to the new argument.
- [x] the `View` widget is now displayed by default. Reasons: the users will almost always want to show the `View` widget so better to show it by default.

### DIFF
--- a/src/napari_spatialdata/__main__.py
+++ b/src/napari_spatialdata/__main__.py
@@ -62,9 +62,7 @@ def view(path: tuple[str], headless: bool) -> None:
         break
 
     # TODO: remove [0] when multiple datasets are supported
-    interactive = Interactive(sdata=sdatas[0])
-    if not headless:
-        interactive.run()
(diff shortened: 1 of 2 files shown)


melonora (approved):
Thanks for the PR. One question, but not blocking right now. Should we change the notebooks to use `Interactive` in headless?

LucaMarconato:
In the notebooks, for testing reasons, I would go for that (we could use an environmental variable that flags that we are testing the notebooks). I would also use the approach described here https://github.com/scverse/napari-spatialdata/issues/123 to make the napari instance test the whole set of layers.

PR #144, opened 2023-07-31
Test loading elements without viewer

rahulbshrestha:
This PR introduces the ability to add elements to the Interactive class without opening the Napari viewer. 

Other minor changes:
- The previous `test_interactive.py` file has been renamed to `test_widgets.py`, and the new tests have been added to `test_interactive.py`.


src/napari_spatialdata/_interactive.py
```diff
@@ -27,6 +28,27 @@ class Interactive:
     None
     """
 
+    def add_element(self, coordinate_system_name: str, elements: str) -> SpatialData:
```
LucaMarconato, line 31: rename to `element`

src/napari_spatialdata/utils/_test_utils.py
```diff
@@ -52,3 +58,33 @@ def click_list_widget_item(qtbot: QtBot, widget: QListWidget, position: QPoint,
...
+
+    return interactive_screenshot
+
+
+def save_image(image_np: NDArrayA, file_name: str) -> None:
+    """Save image to file. This was used to generate tests/plots/plot_image.png.
+
+    Parameters
+    ----------
+    image_np: NDArrayA
```
LucaMarconato, line 86: same here with `: NDArrayA`

tests/plots/groundtruth/plot_image.tiff
LucaMarconato, line 1: why tiff and not png? I'd remove this as it is heavy and do a comparison without equality but with a certain tolerance (this is also done in spaitaldata-plot and squidpy)

tests/test_interactive.py
```diff
@@ -1,287 +1,19 @@
...
-    widget = widget(viewer)
-    assert widget.model.adata is adata_labels
-    viewer.add_image(
-        image,
-        rgb=True,
-        name="image",
-        metadata={"adata": adata_shapes},
-    )
-    assert widget.model.adata is adata_shapes
+    assert np.array_equal(screenshot, presaved_screenshot)
```
LucaMarconato, line 19: very good, it if works even with `np.array_equal()`, even better.

tests/plots/groundtruth/plot_image.tiff
LucaMarconato: Indeed, this file is not used in the test and it can be removed.

LucaMarconato (approved):
Only minor changes requested, very good.

melonora (commented):
Overall looks good, but I would provide the additional option of either showing just the canvas or to also show the viewer.

src/napari_spatialdata/utils/_test_utils.py
```diff
@@ -52,3 +58,40 @@ def click_list_widget_item(qtbot: QtBot, widget: QListWidget, position: QPoint,
...
+    viewer
+        Instance of napari Viewer.
+
+    canvas_only
+        If True, only the canvas is saved, not the viewer window.
+    """
+    logger.info("Taking screenshot of viewer")
+    # to distinguish between the black of the image background and the black of the napari background (now white)
+    # TODO (melonora): remove when napari allows for getting rid of margins.
+    viewer.theme = "light"
```
LucaMarconato, line 77: To be sure to restore the old state it is better to do:
old_theme = viewer.theme
viewer.theme = "light"
# take screenshot
viewer.theme = old_theme

src/napari_spatialdata/_interactive.py
```diff
@@ -27,6 +28,27 @@ class Interactive:
...
+                elif element_type == "shapes":
+                    if type(self._sdata.shapes[element].iloc[0][0]) == shapely.geometry.point.Point:
+                        self._sdata_widget.viewer_model.add_sdata_circles(self._sdata, coordinate_system_name, element)
+                    elif (type(self._sdata.shapes[element].iloc[0][0]) == shapely.geometry.polygon.Polygon) or (
+                        type(self._sdata.shapes[element].iloc[0][0]) == shapely.geometry.multipolygon.MultiPolygon
+                    ):
+                        self._sdata_widget.viewer_model.add_sdata_shapes(self._sdata, coordinate_system_name, element)
+
+                break
+
```
LucaMarconato, line 51: thinking better about this, I think it's not good that we are duplicating the logic of _sdata_widgets.py/_add_shapes(), 

I would remove this code and the corresponding code from _add_shapes(); put it in a function; and call that function from here and from _add_shapes(). The function should tell if we are dealing with a point or with a polygon/multipolygon. It could for instance returnt the type directly
```suggestion
            if element_name == element:
                if element_type == "images":
                    self._sdata_widget.viewer_model.add_sdata_image(self._sdata, coordinate_system_name, element)
                elif element_type == "labels":
                    self._sdata_widget.viewer_model.add_sdata_labels(self._sdata, coordinate_system_name, element)
                elif element_type == "points":
                    self._sdata_widget.viewer_model.add_sdata_points(self._sdata, coordinate_system_name, element)
                elif element_type == "shapes":
                    if type(self._sdata.shapes[element].iloc[0][0]) == shapely.geometry.point.Point:
                        self._sdata_widget.viewer_model.add_sdata_circles(self._sdata, coordinate_system_name, element)
                    elif (type(self._sdata.shapes[element].iloc[0][0]) == shapely.geometry.polygon.Polygon) or (
                        type(self._sdata.shapes[element].iloc[0][0]) == shapely.geometry.multipolygon.MultiPolygon
                    ):
                        self._sdata_widget.viewer_model.add_sdata_shapes(self._sdata, coordinate_system_name, element)

                break

```

LucaMarconato, line 51: thinking better about this, I think it's not good that we are duplicating the logic of _sdata_widgets.py/_add_shapes(), 

I would remove this code and the corresponding code from _add_shapes(); put it in a function; and call that function from here and from _add_shapes(). The function should tell if we are dealing with a point or with a polygon/multipolygon. It could for instance returnt the type directly
```suggestion
            if element_name == element:
                if element_type == "images":
                    self._sdata_widget.viewer_model.add_sdata_image(self._sdata, coordinate_system_name, element)
                elif element_type == "labels":
                    self._sdata_widget.viewer_model.add_sdata_labels(self._sdata, coordinate_system_name, element)
                elif element_type == "points":
                    self._sdata_widget.viewer_model.add_sdata_points(self._sdata, coordinate_system_name, element)
                elif element_type == "shapes":
                    if type(self._sdata.shapes[element].iloc[0][0]) == shapely.geometry.point.Point:
                        self._sdata_widget.viewer_model.add_sdata_circles(self._sdata, coordinate_system_name, element)
                    elif (type(self._sdata.shapes[element].iloc[0][0]) == shapely.geometry.polygon.Polygon) or (
                        type(self._sdata.shapes[element].iloc[0][0]) == shapely.geometry.multipolygon.MultiPolygon
                    ):
                        self._sdata_widget.viewer_model.add_sdata_shapes(self._sdata, coordinate_system_name, element)

                break

```

src/napari_spatialdata/_interactive.py
```diff
@@ -27,6 +28,27 @@ class Interactive:
...
+                elif element_type == "points":
+                    self._sdata_widget.viewer_model.add_sdata_points(self._sdata, coordinate_system_name, element)
+                elif element_type == "shapes":
+                    if type(self._sdata.shapes[element].iloc[0][0]) == shapely.geometry.point.Point:
+                        self._sdata_widget.viewer_model.add_sdata_circles(self._sdata, coordinate_system_name, element)
+                    elif (type(self._sdata.shapes[element].iloc[0][0]) == shapely.geometry.polygon.Polygon) or (
+                        type(self._sdata.shapes[element].iloc[0][0]) == shapely.geometry.multipolygon.MultiPolygon
+                    ):
+                        self._sdata_widget.viewer_model.add_sdata_shapes(self._sdata, coordinate_system_name, element)
+
```
LucaMarconato, line 49: thinking better about this, I think it's not good that we are duplicating the logic of _sdata_widgets.py/_add_shapes(),

I would remove this code and the corresponding code from _add_shapes(); put it in a function; and call that function from here and from _add_shapes(). The function should tell if we are dealing with a point or with a polygon/multipolygon. It could for instance returnt the type directly

LucaMarconato:
@rahulbshrestha I reviewed the code; it's almost good to merge, but I would suggest some changes (shown in the comments above).

Before merging please complete this checklist:
- [ ] implement the suggestions in the conversations above until all are marked as resolved
- [ ] merge main into this branch
- [ ] fix the failing tests

melonora (changes requested):
Thanks for the PR! I need to test this pr more thoroughly and also have some idea about getting rid of the duplicated logic, but would have to test first. In any case one addition to the tests I think would be useful is when adding multiple elements in different coordinate spaces. This might also require us to first answer the question what we think should be the expected logic here.

src/napari_spatialdata/_interactive.py
LucaMarconato: The check about the first element of the series is more like a temporary hack, and we should replace it with what is being done here: https://github.com/scverse/spatialdata/blob/f58c4f426ab75d070558cf7ec1cb7c03b780d1fa/src/spatialdata/_core/query/_utils.py#L4, so bundling it into a function is more to set a placeholder for the new code, and avoid risking forgetting to update both parts of the code when we apply this change.

LucaMarconato: Actually, let's keep the function as it is now, gonna refactor it in `spatialdata` and then make a PR to this repo.

LucaMarconato:
> 

- [ ] I agree, I'd add the option to set the coordinate space when calling interactive, it can be an extra parameter. 

After this we can merge.

LucaMarconato (approved):

PR #146, opened 2023-08-12
Transform layers

melonora:
This PR allows for the affine transform associated with the layer(s) to be updated when the coordinate system is changed. It will ignore layers that do not have an `sdata` object associated with them. 

Furthermore, this PR includes a bugfix for the issue of  a `KeyError` being raised when the visibility of the layer is updated and the layer has no `sdata` associated with it.

### DIFF
--- a/src/napari_spatialdata/utils/_utils.py
+++ b/src/napari_spatialdata/utils/_utils.py
@@ -179,7 +179,7 @@ def _swap_coordinates(data: list[Any]) -> list[Any]:
 
 
 def _get_transform(element: SpatialElement, coordinate_system_name: str | None = None) -> None | NDArrayA:
-    if not isinstance(element, (SpatialImage, MultiscaleSpatialImage, AnnData, DaskDataFrame, GeoDataFrame)):
+    if not isinstance(element, (SpatialImage, MultiscaleSpatialImage, DaskDataFrame, GeoDataFrame)):
         raise RuntimeError("Cannot get transform for {type(element)}")
 
     transformations = get_transformation(element, get_all=True)
--- a/src/napari_spatialdata/utils/_utils.py
+++ b/src/napari_spatialdata/utils/_utils.py
@@ -179,8 +179,6 @@ def _swap_coordinates(data: list[Any]) -> list[Any]:
 
 
 def _get_transform(element: SpatialElement, coordinate_system_name: str | None = None) -> None | NDArrayA:
-    affine: NDArrayA
-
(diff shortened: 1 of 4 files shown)


src/napari_spatialdata/utils/_utils.py
```diff
@@ -178,25 +178,16 @@ def _swap_coordinates(data: list[Any]) -> list[Any]:
...
 
-def _get_transform(element: SpatialElement, coordinate_system_name: str | None = None) -> NDArrayA:
-    affine: NDArrayA
-    transformations = get_transformation(element, get_all=True)
-    cs = transformations.keys().__iter__().__next__() if coordinate_system_name is None else coordinate_system_name
-    ct = transformations[cs]
-    affine = ct.to_affine_matrix(input_axes=("y", "x"), output_axes=("y", "x"))
-
+def _get_transform(element: SpatialElement, coordinate_system_name: str | None = None) -> None | NDArrayA:
     if not isinstance(element, (SpatialImage, MultiscaleSpatialImage, AnnData, DaskDataFrame, GeoDataFrame)):
```
LucaMarconato, line 182: I removed AnnData, it could have transformations associated to it only in a very old version of the SpatialData class

src/napari_spatialdata/utils/_utils.py
```diff
@@ -178,25 +178,16 @@ def _swap_coordinates(data: list[Any]) -> list[Any]:
...
         raise RuntimeError("Cannot get transform for {type(element)}")
-    # elif isinstance(element, (AnnData, DaskDataFrame, GeoDataFrame)):
-    #    return affine
-    #    from spatialdata.transformations import Affine, Sequence, MapAxis
-    #    new_affine = Sequence(
```
LucaMarconato, line 193: Are you sure this is not needed? If so, better since I didn't like it, but I had to add this swap of coordiantes to make things work (I don't remember if in the context of displaying thigns properly, or of saving properly the annotatoins)

melonora: hmm would be worth to double check

melonora: At least with applying translation it wasn't

LucaMarconato: true also checked, it works

src/napari_spatialdata/utils/_utils.py
LucaMarconato: Ah you removed before me :P

LucaMarconato (approved):
Looks very good to me! Tried on some datasets and it works. Nice fix!

PR #148, opened 2023-08-18
Fix _current deprecation

### LINKED ISSUE #137: Selection._current deprecation napari 0.5.0
Here and there we make use of `Selection._current`. This way of accessing will become deprecated with napari 0.5.0 and should be changed.

melonora:
Closes #137 
From napari 0.5.0 onwards the use of `Selection._current` from a plugin would become deprecated. This currently gives warnings. The alternative would be to use `Selection.active`. The behaviour of this is slightly different from `Selection._current`, but this does not affect our code.

In short the difference in behaviour is that when having a `Selection` of multiple layers, there is no `Selection.active`, while there can be `Selection._current`.

## Reference
https://napari.org/stable/api/napari.utils.events.Selection.html


LucaMarconato:
Thank you, very nice!

giovp:
very nice indeed! I would merge #136 first and see if the diffs look better?

melonora:
yeah I will rebase once #136 is merged

PR #152, opened 2023-08-23
Switch to currentItemChanged

### LINKED ISSUE #103: Split click callbacks in click and on change
From my understanding most of the callbacks are being triggered by clicks instead of item changes. This would mean that programmatically if you change an item no callback is necessarily triggered. I think we have to discuss separating callbacks where the click callbacks would only change the row for example in a widget which then triggers another callback actually changing anything that is displayed. For example:

1. coordinate_system widget on click on another coordinate system change the current coordinate system.
2. on coordinate system change update layer visibility.

What do you think? I think this is also easier maintainable when writing tests as right now I believe we have to simulate mouseclicks every time.

### LINKED ISSUE #127: Support changing coordinate system with arrow keys
Single clicks on the coordinate systems list widget changes the coordinate systems. I think it will be handy that also navigating with the arrow keys would lead to the same behavior. Should be a quick fix.

### LINKED ISSUE #128: Single click instead of double click for elements widget?
The coordinate system widget operates with single clicks. Contrariwise, the elements widget requires a double click to add a new element. What's the reason for this?

I think it would be more homogeneous to make the elements widget react to single click (and eventually also arrow keys, analogously to what described in this issue https://github.com/scverse/napari-spatialdata/issues/127).

melonora:
Closes #103 
Closes #127 
Closes #128
For the `CoordinateSystemWidget` the only way to change was by clicking. This PR introduces a change so that the callbacks happen on `currentItemChanged`. This item can change based on click but also when using the arrow keys.


melonora:
I will rebase this one to main once #136 is merged. One question, should we have an `event` being emitted when `CoordinateSystemWidget._coordinate_system` changes?

LucaMarconato:
Very nice thanks! I think the event is not needed atm so I would skip it; maybe in the future we will need this and we can add this modification.

Issue #150, opened 2023-08-22
Non-RGB multichannels can be 3


aeisenbarth:
In pull request #115 for issue #66, a logic was implemented that converts multichannels to RGB if they are three.

In our use case, we have non-RGB channels, that can be 2, 4 or sometimes 3. The problem is that independent channels have very different value ranges and are not necessarily normalized, so that one RGB component is dominating. Furthermore, the visualization does not represent physical colors and can be misleading to the viewer.

![image](https://github.com/scverse/napari-spatialdata/assets/54448967/c796bda4-91ca-4b3d-b3a7-d8b3b5be80b4)

Is there a way to set up the SpatialData object to explicitly use colors from the metadata, and avoid interpretation as RGB?

I think Napari's logic is that `c` must be 3 and additionally the order provided by the user must be `yxc` to interpret an image as RGB.

melonora:
Hi @aeisenbarth, at the moment it is the case that in case of 3 or 4 channels it is interpreted as RGB. I will implement an RGB argument similar to napari which should support your usecase.

aeisenbarth:
I also found that in [_get_affine_for_images_labels](https://github.com/scverse/napari-spatialdata/blob/2590e1e8191091820478b470ef7aee8bfd322be8/src/napari_spatialdata/interactive.py#L120) (on branch spatialdata), we could also inspect channel names, if these are always available for true RGB images:
```python
{"r", "g", "b"} <= set(element["scale0"].coords.indexes["c"]) # for MultiscaleSpatialImage
```

melonora:
These are not always available for true RGB images as far as I know. @LucaMarconato could you pitch in?

LucaMarconato:
Thanks @aeisenbarth for reporting. I think that a quick workaround would be to implement the RGB argument mentioned by @melonora to mimic the napari's functionality. I also think that the clean solution would be to rely exclusively on channel name and ensure that they are always present when needed by:
1) add an argument `rgb: bool = None` in `ImageModel.parse()` to give a convenience function to the user to set the channel names. The behavior could be to set the names to `r`, `g`, `b` based on the number of channels when `rgb` is `None` or is `True`, but also to give a printed log to the user when `rgb` is `None`.
2) modifying the readers of spatialdata-io to use the `rgb=True` parameter when calling `ImageModel.parse()`.

aeisenbarth:
I could do a PR for this.

[...]

PR #153, opened 2023-08-24
Change RGB detection to be based on channel names

aeisenbarth:
See https://github.com/scverse/napari-spatialdata/issues/150

To give users more control over whether a 3/4-channel image is interpreted and visualized as RGB, channel names now need to be explicitly set to ["r", "g", "b"]. Therefore, [`to_spatial_image`](https://github.com/spatial-image/spatial-image/pull/21) would have a new argument `rgb=True` to override the channel names (or auto-detect with `rgb=None`).


aeisenbarth:
Since the condition `{"r", "g", "b"} <= set(c_coords)` does not require a specific order of `c_coords`, I should also reorder them to RGB, since that is what Napari will use. Does anyone know the corresponding `xarray` function (analog to transposing dimensions)?

aeisenbarth:
Tests fail because multiscales blobs are not correctly set up with provided channel names. https://github.com/scverse/spatialdata/pull/342 would need to be merged first.

melonora:
@aeisenbarth Thanks again, sorry it took some time

PR #155, opened 2023-08-30
Refactor duplicate logic

melonora:
Currently, on main there is a lot of duplicated code. This PR provides a refactor to remove duplicated logic in `_sdata_widgets` and `Interactive`.


melonora:
Just to note, I added the `_elements` and `_system` in the respective `__init__` of the classes because otherwise `_onClick` could not be directly called to prevent duplicated logic. Perhaps there are other ideas to do this, but for now, I did it like this.

LucaMarconato (commented):
gonna change this

LucaMarconato (approved):

LucaMarconato:
Much cleaner code now, great work!

PR #168, opened 2023-10-15
Saving / exporting new elements back to SpatialData

### LINKED ISSUE #83: Support saving shapes / points when visualizing multiple spatialdata objects at once
Currently on main it is not supported to visualize multiple spatialdata objects, but this will be supported in the future. Once supported on the main branch, it should be supported to save shapes / points to multiple spatialdata objects at the same time.

### LINKED ISSUE #160: Save back to zarr
I am following the [landmark annotation tutorial](https://spatialdata.scverse.org/en/latest/tutorials/notebooks/notebooks/examples/alignment_using_landmarks.html) to annotate points in a SpatialData object. 

However, I'm working on a remote system without a graphical user interface, so I can't use `Interactive()`. But I have napari available on my local machine and can open the `.zarr` files. 

This works reasonably well, however I don't seem to be able to save the modifications back to the zarr store such that I can use the annotations from my ipython session. Is this possible somehow? 

I believe this would also be very useful for scientists who don't use python to analyze data (e.g. pathologists). Like that I could provide them with a link to the zarr object and they could just annotate & save.

melonora:
This PR allows for saving / exporting back new layers to the SpatialData zarr store. Furthermore, it adds preparation code for being able to make changes in those layers visualizing existing elements in a SpatialData object and storing those back. 
A requirement for saving new layers / elements is to first link it to an existing SpatialData object by `shift+click` on the new layer and a layer with the already existing SpatialData object and pressing `shift+L`. Afterwards select the layer to be saved and press `shift+S`.

The PR is dependent on [#329](https://github.com/scverse/spatialdata/pull/329) for partially updating SpatialData objects in memory. A follow up PR of this one will allow for writing partial updates back to disk.

*EDIT from Luca: users can use [this other (legacy) branch](https://github.com/scverse/napari-spatialdata/tree/spatialdata) to have the functionality enabled before the current PR is merged.
The current PR gives a more stable implementation, has more tests and deals with the case of multiple `SpatialData` objects in a transparent way, letting the user choose where to save the annotations*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,3 +113,6 @@ spatialdata-sandbox
 
 # Plots for testing
 tests/plots/generated
+
+# Local temp script for testing user bugs (Luca)
+temp/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,7 +49,7 @@ repos:
     hooks:
       - id: blacken-docs
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.1.3
+    rev: v0.1.4
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,13 +7,13 @@ default_stages:
 minimum_pre_commit_version: 2.9.3
 repos:
   - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.5.1
+    rev: v1.6.1
     hooks:
       - id: mypy
         additional_dependencies: [numpy>=1.23]
         exclude: docs
   - repo: https://github.com/psf/black
-    rev: 23.9.1
+    rev: 23.10.1
     hooks:
       - id: black
         additional_dependencies: [toml]
@@ -49,7 +49,7 @@ repos:
     hooks:
       - id: blacken-docs
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.0.292
+    rev: v0.1.3
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -14,3 +14,4 @@ python:
       extra_requirements:
         - "doc"
         - "all"
+        - "readthedocs"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,19 @@ and this project adheres to [Semantic Versioning][].
 [keep a changelog]: https://keepachangelog.com/en/1.0.0/
 [semantic versioning]: https://semver.org/spec/v2.0.0.html
 
-## [0.2.8] - tbd
+## [0.2.9] - TBD
+
+## [0.2.8] - 2023-10-30
+
+### Added
+
+- Global parameters POINT_THRESHOLD and POLYGON_THRESHOLD added to allow users to adjust these parameters.
+
+### Fixed
+
+- Updated dependencies
+- Updated installation instructions since napari conda-forge no longer installs Qt backend. @psobolewskiPhD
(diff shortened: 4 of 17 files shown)


LucaMarconato:
@melonora I think we had the same error in building the docs last week. How did you fix it?

melonora:
In that case it was that the branch was not updated yet to use the proper versions.

melonora:
Btw I think you checked that time on the denbi machine.

LucaMarconato:
Ah true it's the same bug, ok gonna check what's going on.

LucaMarconato:
Found the problem, `ActionType` is not available in the current `napari`. Is it this one that requires `napari==0.4.19`? If so, do you know when it will be released?

Btw for debugging this the denbi machine was not needed, I used this trick: https://github.com/scverse/napari-spatialdata/pull/168/commits/b5af124c4ff6619cc636613ab6136f0e30991a16

melonora:
> Found the problem, `ActionType` is not available in the current `napari`. Is it this one that requires `napari==0.4.19`? If so, do you know when it will be released?
> 
> Btw for debugging this the denbi machine was not needed, I used this trick: [b5af124](https://github.com/scverse/napari-spatialdata/commit/b5af124c4ff6619cc636613ab6136f0e30991a16)

The release candidate is already out, so I expect in 1-2 weeks. As for `ActionType` that is weird, see here on the 0.4.18 branch: [code](https://github.com/napari/napari/blob/20cf5b2c7bcdad26adffa24c778b660cf89483e1/napari/layers/base/__init__.py#L1-L4)

melonora:
> Found the problem, `ActionType` is not available in the current `napari`. Is it this one that requires `napari==0.4.19`? If so, do you know when it will be released?
> 
> Btw for debugging this the denbi machine was not needed, I used this trick: [b5af124](https://github.com/scverse/napari-spatialdata/commit/b5af124c4ff6619cc636613ab6136f0e30991a16)

The release candidate is already out, so I expect in 1-2 weeks. As for `ActionType` that is weird, see here on the 0.4.18 branch: [code](https://github.com/napari/napari/blob/20cf5b2c7bcdad26adffa24c778b660cf89483e1/napari/layers/base/__init__.py#L1-L4)

LucaMarconato:
@melonora 

- I have implemented the behavior that I described in the docstring (when only one sdata is present, save directly with Shift+E and do the linking automatically; when > 1 sdata are present tell the user how to link layers with Shift+L).
- There was a bug that required the adjustment of coordinates (see GIF). I fixed it with `data_to_world()`. In the GIF I am doing the following:
	- Show an image, which has an `Identity` transformation
	- create and save some shapes
	- refresh the coordinate system
	- a new layer with the saved shapes is loaded in the correct location
	- Now show the shapes, which have a non-identical `Affine` transformation
	- create and save some shapes
	- refresh the coordinate system
	- Now the shapes are recreated in the wrong place

There are two more things to do and then we can merge:
- [ ] I added a TODO for refreshing the coordinate system widget after an element is saved, can you help me with that please?
- [ ] I need to add tests for the new implementation:
  - [ ] save when nothing is selected (should throw an error)
  - [ ] save when something with already a name is selected (should throw an error)
  - [ ] save with selection > 1 (should throw an error)
  - [ ] save with multiple spatialdata object (should throw an error)
  - [ ] save with multiple spatialdata objects after having linked (should work)
  - [ ] test both points and shapes; saving, reloading, resaving should lead to the same coordinates

I will implement the tests in the afternoon, if you have time to start implementing them it would be appreciated, but no problem if you can't.

CleanShot 2023-11-09 at 03 39 41

LucaMarconato:
@melonora I have implemented the missing feature and the tests, ready for review!

melonora:
@LucaMarconato I cannot approve my own pull request hehe. However, I did have a look and it seems we are good to go. I just added one comment as to why we have the print statement in `conf.py`. Feel free to approve and merge.

LucaMarconato:
Amazing! Thanks a lot!

LucaMarconato (approved):

PR #169, opened 2023-10-15
Public API switch coordinate system

melonora:
This PR adds the opportunity to change the coordinate system through a public API, thereby completing one task in #158.

PR #170, opened 2023-10-16
Add element api

### LINKED ISSUE #159: Cannot plot categorical annotations on circles
Code to reproduce.

The code does the following:
1. adapts a `blobs` sdata object to have a table, with categorical obs, annotating the circles;
2. plots the obs column `annotation` onto the `blobs_circles` element with `spatialdata-plot`, plot shown below;
3. if you try to plot the same obs column with napari you get an error, reported below.

```python
import spatialdata as sd
from spatialdata.models import TableModel
from spatialdata.datasets import blobs
import spatialdata_plot
import numpy as np
import matplotlib.pyplot as plt
import matplotlib.patches as patches

sdata = blobs()

# let's fetch the table metadata
region = sdata.table.uns[TableModel.ATTRS_KEY]['region']
region_key = sdata.table.uns[TableModel.ATTRS_KEY]['region_key']
instance_key = sdata.table.uns[TableModel.ATTRS_KEY]['instance_key']

# let's see what the table annotates
print(region)

# it annotates the labels; let's make the table annotate the circles instead
new_table = sdata.table.copy()
sdata.table.obs[region_key] = 'blobs_circles'

# there were 21 labels, but now there are only 5 circles, we need to subset
new_table = sdata.table[:5]
new_table.uns[TableModel.ATTRS_KEY]['region'] = 'blobs_circles'

# double checking that the indices match
print(new_table.obs[instance_key].to_numpy())
print(sdata['blobs_circles'].index.to_numpy())

# they don't, let's fix this
new_table.obs[instance_key] = np.array(range(5))

# let's add some annotations
new_table.obs['annotation'] = ['a', 'b', 'c', 'd', 'e']
new_table.obs['annotation'] = new_table.obs['annotation'].astype('category')

# the new table is ready, let's replace the old one
del sdata.table
sdata.table = new_table

sdata.pl.render_shapes('blobs_circles', color='annotation').pl.show()

Interactive(sdata)
```

Expected result (note, the colors with napari may be different, also the size will be different due to this bug https://github.com/scverse/napari-spatialdata/issues/35#issuecomment-1740897304, that needs to be fixed upstream in napari):
image

Error:
```
2023-09-29 16:02:12.811 | DEBUG    | napari_spatialdata.utils._utils:_get_categorical:146 - KEY: annotation
Traceback (most recent call last):
  File "/Users/macbook/embl/projects/basel/napari-spatialdata/src/napari_spatialdata/_widgets.py", line 64, in <lambda>
    self.itemDoubleClicked.connect(lambda item: self._onAction((item.text(),)))
  File "/Users/macbook/embl/projects/basel/napari-spatialdata/src/napari_spatialdata/_widgets.py", line 139, in _onAction
    properties = self._get_points_properties(vec, key=item, layer=self.model.layer)
  File "/Users/macbook/miniconda3/envs/ome/lib/python3.10/functools.py", line 926, in _method
    return method.__get__(obj, cls)(*args, **kwargs)
  File "/Users/macbook/embl/projects/basel/napari-spatialdata/src/napari_spatialdata/_widgets.py", line 247, in _
    cluster_labels = _position_cluster_labels(self.model.coordinates, vec)
  File "/Users/macbook/embl/projects/basel/napari-spatialdata/src/napari_spatialdata/utils/_utils.py", line 153, in _position_cluster_labels
    coords = coords[:, 1:]
TypeError: 'NoneType' object is not subscriptable
```

melonora:
Closes #159 

This is a follow up PR of #169 and should finish the last task of #159. When adding an element now the element will always be visible even when the element is not present in the current `coordinate_system`. However, if `view_element_system` is set to `True`, it is switched to the `coordinate_system` specified by the user for viewing the element.

### DIFF
--- a/src/napari_spatialdata/_interactive.py
+++ b/src/napari_spatialdata/_interactive.py
@@ -10,6 +10,7 @@ from napari_spatialdata.utils._utils import (
     NDArrayA,
     get_duplicate_element_names,
     get_elements_meta_mapping,
+    get_itemindex_by_text,
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


LucaMarconato:
Look good to me! Nice work!

LucaMarconato (approved):

PR #173, opened 2023-10-22
Improve linking new layers to sdata

melonora:
This PR allows for the user that if a layer with `sdata` in the metadata is created and a new layer is created from the napari menu to link the new layer without having to select the layer with `sdata` in the metadata. In case multiple `sdata` objects are in the opened layers it will check to see if there is only one `sdata` in the metadata of the layers in `layers.selection`. If there is no `sdata` an error gets thrown. If exactly 1 `sdata` the layer will be linked. If more than 1 `sdata` an errow will be thrown.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,7 @@ repos:
     hooks:
       - id: prettier
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.4.0
+    rev: v4.5.0
     hooks:
       - id: detect-private-key
       - id: check-merge-conflict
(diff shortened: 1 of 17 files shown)


melonora:
There seems to be a problem with `pytest.raises` not catching errors in the `Qt` event loop. I am not certain how to make that happen yet.

melonora:
I adjusted the test to just directly call the method. In this way we don't run into problems.

LucaMarconato:
Looks very good, very useful! I adjusted some docstrings and I split one function to improve a bit clarity and allow for code reuse. Let's see if the CI passes (tests pass locally). Then for me you can merge!

LucaMarconato (approved):

PR #175, opened 2023-10-26
Plotting annotations fix

### LINKED ISSUE #164: Error when plotting annotations when the element "region" is not selected
### The bug
In the blobs dataset, the table has var names `'channel_0_sum', 'channel_1_sum', 'channel_2_sum'`, associated to the labels.
Plotting any of these values on the labels using napari (the code to run napari is at the end), works.
Unfortunately, if one the selected another element (circles, points, polygons, multipolygons, multiscale_labels), and tries to plot the var annotations, this exception occurs:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
File ~/embl/projects/basel/napari-spatialdata/src/napari_spatialdata/_widgets.py:64, in ListWidget.__init__.<locals>.<lambda>(item=<PyQt6.QtWidgets.QListWidgetItem object>)
     61 self._unique = unique
     62 self._viewer = viewer
---> 64 self.itemDoubleClicked.connect(lambda item: self._onAction((item.text(),)))
        self = <napari_spatialdata._widgets.AListWidget object at 0x285d01510>
        item = <PyQt6.QtWidgets.QListWidgetItem object at 0x2c99471c0>
     65 self.enterPressed.connect(self._onAction)
     66 self.indexChanged.connect(self._onAction)

File ~/embl/projects/basel/napari-spatialdata/src/napari_spatialdata/_widgets.py:139, in AListWidget._onAction(self=<napari_spatialdata._widgets.AListWidget object>, items=('channel_1_sum',))
    130     self.viewer.add_points(
    131         vec,
    132         name=name,
   (...)
    136         symbol=self.model.symbol,
    137     )
    138 else:
--> 139     properties = self._get_points_properties(vec, key=item, layer=self.model.layer)
        item = 'channel_1_sum'
        vec = <class 'numpy.ndarray'> (0,) float64
        self = <napari_spatialdata._widgets.AListWidget object at 0x285d01510>
    141     if isinstance(self.model.layer, (Points, Shapes)):
    142         self.model.layer.name = (
    143             "" if self.model.system_name is None else self.model.layer.name + ":"
    144         ) + item

File ~/miniconda3/envs/ome/lib/python3.10/functools.py:926, in singledispatchmethod.__get__.<locals>._method(*args=(<class 'numpy.ndarray'> (0,) float64,), **kwargs={'key': 'channel_1_sum', 'layer': <Shapes layer 'blobs_multipolygons'>})
    924 def _method(*args, **kwargs):
    925     method = self.dispatcher.dispatch(args[0].__class__)
--> 926     return method.__get__(obj, cls)(*args, **kwargs)
        method = <function AListWidget._ at 0x15b65f370>
        obj = <napari_spatialdata._widgets.AListWidget object at 0x285d01510>
        cls = <class 'napari_spatialdata._widgets.AListWidget'>
        args = (<class 'numpy.ndarray'> (0,) float64,)
        kwargs = {'key': 'channel_1_sum', 'layer': <Shapes layer 'blobs_multipolygons' at 0x2c9e6ead0>}

File ~/embl/projects/basel/napari-spatialdata/src/napari_spatialdata/_widgets.py:203, in AListWidget._(self=<napari_spatialdata._widgets.AListWidget object>, vec=<class 'numpy.ndarray'> (0,) float64, **kwargs={'key': 'channel_1_sum'})
    201 layer = kwargs.pop("layer", None)
    202 cmap = plt.get_cmap(self.model.cmap)
--> 203 norm_vec = _min_max_norm(vec)
        vec = <class 'numpy.ndarray'> (0,) float64
    204 color_vec = cmap(norm_vec)
    205 if layer is not None and isinstance(layer, Labels):

File ~/embl/projects/basel/napari-spatialdata/src/napari_spatialdata/utils/_utils.py:174, in _min_max_norm(vec=<class 'numpy.ndarray'> (0,) float64)
    171 if vec.ndim != 1:
    172     raise ValueError(f"Expected `1` dimension, found `{vec.ndim}`.")
--> 174 maxx, minn = np.nanmax(vec), np.nanmin(vec)
        vec = <class 'numpy.ndarray'> (0,) float64
        np = <module 'numpy' from '/Users/macbook/miniconda3/envs/ome/lib/python3.10/site-packages/numpy/__init__.py'>
    176 return (  # type: ignore[no-any-return]
    177     np.ones_like(vec) if np.isclose(minn, maxx) else ((vec - minn) / (maxx - minn))
    178 )

File <__array_function__ internals>:200, in nanmax(*args=(<class 'numpy.ndarray'> (0,) float64,), **kwargs={})

File ~/miniconda3/envs/ome/lib/python3.10/site-packages/numpy/lib/nanfunctions.py:476, in nanmax(a=<class 'numpy.ndarray'> (0,) float64, axis=None, out=None, keepdims=<no value>, initial=<no value>, where=<no value>)
    471     kwargs['where'] = where
    473 if type(a) is np.ndarray and a.dtype != np.object_:
    474     # Fast, but not safe for subclasses of ndarray, or object arrays,
    475     # which do not implement isnan (gh-9009), or fmax correctly (gh-8975)
--> 476     res = np.fmax.reduce(a, axis=axis, out=out, **kwargs)
        kwargs = {}
        np = <module 'numpy' from '/Users/macbook/miniconda3/envs/ome/lib/python3.10/site-packages/numpy/__init__.py'>
        a = <class 'numpy.ndarray'> (0,) float64
        np.fmax = <ufunc 'fmax'>
        axis = None
        out = None
    477     if np.isnan(res).any():
    478         warnings.warn("All-NaN slice encountered", RuntimeWarning,
    479                       stacklevel=3)

ValueError: zero-size array to reduction operation fmax which has no identity
WARNING: skipping QEventPoint(id=1 ts=0 pos=0,0 scn=2431.49,1158.77 gbl=2431.49,1158.77 Released ellipse=(1x1 ∡ 0) vel=0,0 press=-2431.49,-1158.77 last=-2431.49,-1158.77 Δ 2431.49,1158.77) : no target window
```

### Expected behavior
The expected behavior would be to prevent the plot to be made by clearing the table widget on the right. In fact, the newly selected elements have no table associated, so the user should have not had the option to plot `channel_0_sum`. Instead, the user should only see what can be plotted on top of the element (e.g. columns of the dataframe).

### Solution
I suggested to implement the following:
- [x] updating the widgets on the right with the correct content every time an element is selected
- [ ] adding tests for the above for all the elements of `blobs()`

### Code to open napari.
```python
from spatialdata.datasets import blobs
Interactive(blobs())
```

melonora:
closes #164 

This PR fixes the issue reported in #164. The problem was that when filtering the anndata object it would result in an anndata object with potentially 0 rows. This PR checks whether the anndata object to be added to the metadata has 0 rows. If this is the case then the `adata` metadata is set to be `None`. Also if there is no table, adata will be none, but in addition `region_key` will also be `None`.

In addition to the implementation of @aeisenbarth this handles the case when a table is present, but there are no annotations for a given element.

### DIFF
--- a/src/napari_spatialdata/_view.py
+++ b/src/napari_spatialdata/_view.py
@@ -252,12 +252,6 @@ class QtAdataViewWidget(QWidget):
                 self.model.adata.obsm[Key.obsm.spatial][:, ::-1][:, :2], 0, values=0, axis=1
             )
 
-        if "points" in layer.metadata:
-            # TODO: Check if this can be removed
-            self.model.points_coordinates = layer.metadata["points"].X
-            self.model.points_var = layer.metadata["points"].obs["gene"]
-            self.model.point_diameter = np.array([0.0] + [layer.metadata["point_diameter"]] * 2) * self.model.scale
-
         self.model.spot_diameter = np.array([0.0, 10.0, 10.0])
         self.model._labels_key = layer.metadata["region_key"] if isinstance(layer, Labels) else None
         self.model.system_name = layer.metadata["name"] if "name" in layer.metadata else None
--- a/src/napari_spatialdata/_view.py
+++ b/src/napari_spatialdata/_view.py
@@ -256,8 +256,6 @@ class QtAdataViewWidget(QWidget):
         self.model._labels_key = layer.metadata["region_key"] if isinstance(layer, Labels) else None
         self.model.system_name = layer.metadata["name"] if "name" in layer.metadata else None
 
-        if "colormap" in layer.metadata:
-            self.model.cmap = layer.metadata["colormap"]
         if hasattr(
             self, "obs_widget"
         ):  # to check if the widget has been already initialized, layer update should only be called on layer change
--- a/src/napari_spatialdata/_viewer.py
+++ b/src/napari_spatialdata/_viewer.py
@@ -12,6 +12,7 @@ from napari.utils.notifications import show_info
 
 from napari_spatialdata.utils._utils import (
     _adjust_channels_order,
+    _get_metadata_adata,
     _get_transform,
     _swap_coordinates,
     get_duplicate_element_names,
@@ -152,6 +153,7 @@ class SpatialDataViewer:
         xy = np.array([df.geometry.x, df.geometry.y]).T
         xy = np.fliplr(xy)
         radii = df.radius.to_numpy()
+        adata = _get_metadata_adata(sdata, original_name)
 
         self.viewer.add_points(
             xy,
@@ -161,9 +163,7 @@ class SpatialDataViewer:
             edge_width=0.0,
             metadata={
                 "sdata": sdata,
-                "adata": sdata.table[
-                    sdata.table.obs[sdata.table.uns["spatialdata_attrs"]["region_key"]] == original_name
-                ],
+                "adata": adata,
                 "region_key": sdata.table.uns["spatialdata_attrs"]["region_key"],
                 "name": original_name,
                 "_active_in_cs": {selected_cs},
@@ -198,6 +198,7 @@ class SpatialDataViewer:
                 polygons.append(list(df.geometry.iloc[i].exterior.simplify(tolerance=2).coords))
         # this will only work for polygons and not for multipolygons
         polygons = _swap_coordinates(polygons)
+        adata = _get_metadata_adata(sdata, key)
 
         self.viewer.add_shapes(
             polygons,
@@ -206,7 +207,7 @@ class SpatialDataViewer:
             shape_type="polygon",
             metadata={
                 "sdata": sdata,
-                "adata": sdata.table[sdata.table.obs[sdata.table.uns["spatialdata_attrs"]["region_key"]] == key],
+                "adata": adata,
                 "region_key": sdata.table.uns["spatialdata_attrs"]["region_key"],
                 "name": original_name,
                 "_active_in_cs": {selected_cs},
@@ -221,6 +222,7 @@ class SpatialDataViewer:
 
         affine = _get_transform(sdata.labels[original_name], selected_cs)
         rgb_labels, _ = _adjust_channels_order(element=sdata.labels[original_name])
+        adata = _get_metadata_adata(sdata, key)
 
         self.viewer.add_labels(
             rgb_labels,
@@ -228,7 +230,7 @@ class SpatialDataViewer:
             affine=affine,
             metadata={
                 "sdata": sdata,
-                "adata": sdata.table[sdata.table.obs[sdata.table.uns["spatialdata_attrs"]["region_key"]] == key],
+                "adata": adata,
                 "region_key": sdata.table.uns["spatialdata_attrs"]["instance_key"],
                 "name": original_name,
                 "_active_in_cs": {selected_cs},
--- a/src/napari_spatialdata/_viewer.py
+++ b/src/napari_spatialdata/_viewer.py
@@ -164,7 +164,7 @@ class SpatialDataViewer:
             metadata={
                 "sdata": sdata,
                 "adata": adata,
-                "region_key": sdata.table.uns["spatialdata_attrs"]["region_key"],
+                "region_key": sdata.table.uns["spatialdata_attrs"]["region_key"] if sdata.table else None,
                 "name": original_name,
                 "_active_in_cs": {selected_cs},
                 "_current_cs": selected_cs,
@@ -208,7 +208,7 @@ class SpatialDataViewer:
             metadata={
                 "sdata": sdata,
                 "adata": adata,
-                "region_key": sdata.table.uns["spatialdata_attrs"]["region_key"],
+                "region_key": sdata.table.uns["spatialdata_attrs"]["region_key"] if sdata.table else None,
                 "name": original_name,
                 "_active_in_cs": {selected_cs},
                 "_current_cs": selected_cs,
@@ -231,7 +231,7 @@ class SpatialDataViewer:
             metadata={
                 "sdata": sdata,
                 "adata": adata,
-                "region_key": sdata.table.uns["spatialdata_attrs"]["instance_key"],
+                "region_key": sdata.table.uns["spatialdata_attrs"]["instance_key"] if sdata.table else None,
                 "name": original_name,
                 "_active_in_cs": {selected_cs},
                 "_current_cs": selected_cs,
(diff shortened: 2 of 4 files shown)


melonora:
I added a test that integrates the sdata_widget and adata view widget. Not certain whether we want to move these in a particular file or whether this is ok.

src/napari_spatialdata/_view.py
```diff
@@ -252,18 +252,10 @@ def _select_layer(self) -> None:
                 self.model.adata.obsm[Key.obsm.spatial][:, ::-1][:, :2], 0, values=0, axis=1
             )
 
-        if "points" in layer.metadata:
```
LucaMarconato, line 255: nice!

src/napari_spatialdata/_viewer.py
```diff
@@ -225,16 +222,15 @@ def add_sdata_labels(self, sdata: SpatialData, key: str, selected_cs: str, multi
...
             rgb_labels,
             name=key,
             affine=affine,
             metadata={
                 "sdata": sdata,
-                "adata": sdata.table[sdata.table.obs[sdata.table.uns["spatialdata_attrs"]["region_key"]] == key]
-                if sdata.table
-                else None,
+                "adata": adata,
                 "region_key": sdata.table.uns["spatialdata_attrs"]["instance_key"] if sdata.table else None,
```
LucaMarconato, line 234: in lines 169 and 214 we have "region_key" on the right hand side, here we have "instance_key", I think this is a bug.

LucaMarconato: Could this `instance_key` be the cause of the bug with reversed indices? See my comment in the discussion page.

src/napari_spatialdata/utils/_utils.py
```diff
@@ -367,3 +367,18 @@ def get_elements_meta_mapping(
                 name_to_add = name
             elements[name] = elements_metadata
     return elements, name_to_add
+
+
+def _get_metadata_adata(sdata: SpatialData, key: str) -> None | AnnData:
```
LucaMarconato, line 372: Nice to have this separation, it makes the code more readable. Since there is a bug with the order of the rows (see my comment in the discussion), we could fix it by replacing this function with `match_table_to_element()` https://github.com/scverse/spatialdata/blob/bf3377fd36f94ef51a233ae71312b1c28e978e6e/src/spatialdata/_core/query/relational_query.py#L134, which returns a table with the correct order.

tests/test_viewer.py
```diff
@@ -82,3 +83,22 @@ def test_layer_transform(qtbot, make_napari_viewer: any):
...
+
+    # Click on `global` coordinate system
+    center_pos = get_center_pos_listitem(widget.coordinate_system_widget, "global")
+    click_list_widget_item(qtbot, widget.coordinate_system_widget, center_pos, "currentItemChanged")
+
+    widget._onClick("blobs_labels")
+    assert viewer.layers[-1].metadata["adata"]
+    assert view_widget.obs_widget.item(0)
+
+    # Filtering adata leads to 0 rows so adata should be set to None
```
LucaMarconato, line 101: nice!

LucaMarconato:
Thanks @melonora for the fix! The bug that I have reported in https://github.com/scverse/napari-spatialdata/issues/164 is gone.

During my review I have anyway spotted two parts of the code to double check. 

One is the use of a `instance_key` when in other equivalent parts of the code `region_key` was used.

The other is that I noticed that the current implementation is not robust against reordering of the rows of the table (as I show in the code below). I think that this can be fixed by replacing the function `_get_metadata_adata()` with a call of `match_table_to_element()` (https://github.com/scverse/spatialdata/blob/bf3377fd36f94ef51a233ae71312b1c28e978e6e/src/spatialdata/_core/query/relational_query.py#L134), which returns a table where the rows are reordered to match the element. Anyway, I can't exclude that the bug is caused by the strange use of `instance_key` that I commented above.

Here is the code to reproduce the behavior. The table is essentially the same, but the two visualizations are different.

```python
##
from spatialdata.datasets import blobs

sdata0 = blobs()
table0 = sdata0.table

table1 = table0.copy()
table1.obs = table1.obs.iloc[::-1]
sdata1 = blobs()
del sdata1.table
sdata1.table = table1

print(table0.obs['instance_id'])
print(table1.obs['instance_id'])
##
Interactive(sdata0)
Interactive(sdata1)
```

`sdata0`:
sdata0

`sdata1`:
sdata1

On this topic, @kevinyamauchi @giovp I think this bug is a symptom that we should drop `region`, `region_key`, `instance_key` altogether and just use the obs indices, and by limiting tables to annotate 0 or 1 spatial element, but not more than 1.

LucaMarconato:
@melonora I have checked if this happens for circles and it seems it doesn't: the two plots are identical when sorting the rows. So *it seems* that the bug is not do to not using `match_table_to_element()` but to the use of `instance_key` where there should be `region_key`. I haven't verified this further yet.

melonora:
I see, will look into it

LucaMarconato:
Still trying to understand the reasons for the behavior observed above. Sharing some initial considerations.
- in `model.py`, we should drop `labels_key` and call it `instance_key` https://github.com/melonora/napari-spatialdata/blob/fa33e6292b668b38ce7a7f2058740c989857ae08/src/napari_spatialdata/_model.py#L268; 
- `labels_key` is indeed used only for labels, as per this line `self.model._labels_key = layer.metadata["region_key"] if isinstance(layer, Labels) else None` from `_view.py`; https://github.com/melonora/napari-spatialdata/blob/fa33e6292b668b38ce7a7f2058740c989857ae08/src/napari_spatialdata/_view.py#L256
- when we call `region_key=.... instance_key...` in line https://github.com/melonora/napari-spatialdata/blob/fa33e6292b668b38ce7a7f2058740c989857ae08/src/napari_spatialdata/_viewer.py#L234C8-L234C8, the error is not on the right hand side, but it is in the function argument, which should be called `instance_key`. Currently the `region_key` value is assigned to `labels_key` in `model.py`.
- There is no place where the `instance_key`/`labels_key` is used to sort the rows of the table, so no surprise that after permuting the rows of the table the plots are wrong. But for some reasons for circles the plots is correct after permuting the rows. I haven't yet understood why.
- We should do a deep cleaning of `_model.py`, it's full of legacy code that makes things messy.

LucaMarconato:
Let's merge this PR as it is, since it fixes the case empty tables, and break the problem in two smaller ones. Let's make a PR for cleaning up `_models.py` and after that let's fix the plotting when the rows of the table are shuffled.

PR #176, opened 2023-10-27
update dependencies

melonora:
This PR updates the dependencies. In particular the spatialdata dependency is updated to incorporate the version change of spatial-image. Furthermore, the typing-extension version dependency corresponds to the spatialdata dependency now.

PR #177, opened 2023-10-30
Add safeguard when table is None

aeisenbarth:
Instantiation of `SpatialData` does not create a table, implying that the table is optional.

A SpatialData object with just (visual) elements but no annotations can still be of interest for viewing in napari-spatialdata, however it tried to load annotations without checking whether a table exists.


LucaMarconato (approved):

LucaMarconato:
Perfect thanks a lot! https://github.com/scverse/napari-spatialdata/pull/176 should fix the failing tests, gonna merge it soon.

PR #185, opened 2023-11-09
fixed requirements to match what's used

LucaMarconato:
Some imports were not in the requirements and some required packages were not used. Fixing this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning][].
 
 ## [0.3.1] - TBD
 
-## [0.3.0] - TBD
+## [0.3.0] - 2023-11-09
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


LucaMarconato:
@melonora I checked locally that the installation, the tests and the docs generation work. I think we can merge.

LucaMarconato:
Actually gonna merge and make a release, not much to review here.

psobolewskiPhD:
Something still looks off to me: `napari-plugin-engine` ?
Looks like this is a npe2 plugin (napari.yaml) to me.

melonora:
you are right @psobolewskiPhD. The hook here was used to make the plugin the default reader for `.zarr` which the user can just do via `preferences/plugins` as far as I understand. Will adjust.

LucaMarconato:
Thanks both, I have made a new release. CC @goanpeca

Issue #181, opened 2023-11-05
The plot should be invariant against reordering of the table rows


LucaMarconato:
As it can be observed when running this code: https://github.com/scverse/napari-spatialdata/pull/175#issuecomment-1793767601

Shuffling the rows of the table changes the plots, while it should not. This happens because currently `instance_key` is ignored. A solution is to call the function `match_table_to_element()` to match the rows of the table to the geometries to display https://github.com/scverse/spatialdata/blob/bf3377fd36f94ef51a233ae71312b1c28e978e6e/src/spatialdata/_core/query/relational_query.py#L134.

Weirdly, the bug appears only for labels and not for circles. The reason is probably the use of the fragile mechanism around `_labels_key` in `_model.py`, reason for which I suggest to first clean up the codebase around the model as described here https://github.com/scverse/napari-spatialdata/issues/180, and then address the current issue.

melonora:
Based on #208 I will close this.

Issue #180, opened 2023-11-05
Clean up legacy code in `_models.py` and related parts


LucaMarconato:
CC @giovp, please comment on this.

`_models.py` contains a lot of legacy code that makes the code fragile. In particular we should remove the following:
- [x] `_scale`, not used anymore now that we have coordinate transformations
- [x] `_spot_diameter`/`_point_diameter`, we should keep one of the two only
- [x] `label_key` (and relative methods, like the getter/setter `labels_key`. This is in reality `instance_key`, and for some reasons it's used only for labels. The `region_key` + `instance_key` logic applies both to labels and shapes (shapes=circles or polygons/multipolygons). I suggest either to add `region_key` and `instance_key` to the model, or not put them in the model at all. I favor the second approach.
- [x] remove `_spatial_key`, not needed any more

In this PR I would also make the following change:
- [x] since we don't need both `_spot_diameter` and `_point_diameter`, let's remove the `points` widget from the bottom right UI; it's never used.

giovp:
yes! yes agree on all the points

melonora:
I can pick this up

LucaMarconato:
More on this:
- `spot_diameter` seems to never be used.

LucaMarconato:
@melonora addressed the point on `labels_key` here: https://github.com/scverse/napari-spatialdata/pull/200/files. Great work!

PR #196, opened 2024-03-03
Multi table

melonora:
Allows the user to switch between tables.

### DIFF
--- a/src/napari_spatialdata/_view.py
+++ b/src/napari_spatialdata/_view.py
@@ -158,6 +158,8 @@ class QtAdataViewWidget(QWidget):
         self.table_name_widget = QComboBox()
         if (table_names := self.model.table_names) is not None:
             self.table_name_widget.addItems(table_names)
+
+        self.table_name_widget.currentTextChanged.connect(self._update_adata)
         self.layout().addWidget(table_label)
         self.layout().addWidget(self.table_name_widget)
 
@@ -223,7 +225,7 @@ class QtAdataViewWidget(QWidget):
         logger.info("Updating layer.")
 
         self.table_name_widget.clear()
-        self.table_name_widget.addItems(self._get_element_annotators(event.sources[0]))
+        self.table_name_widget.addItems(self._get_element_annotators())
         self.adata_layer_widget.clear()
         self.adata_layer_widget.addItem("X", None)
         self.adata_layer_widget.addItems(self._get_adata_layer())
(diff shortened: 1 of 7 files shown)


melonora:
still needs test and refactor

src/napari_spatialdata/_view.py
```diff
@@ -256,12 +270,57 @@ def _select_layer(self) -> None:
...
+        if layer is not None and "adata" in layer.metadata:
+            with self.model.events.adata.blocker():
+                self.model.adata = layer.metadata["adata"]
+
+        if self.model.adata.shape == (0, 0):
+            return
+
+        if "spatial" in self.model.adata.obsm:
+            self.model.coordinates = np.insert(
+                self.model.adata.obsm[Key.obsm.spatial][:, ::-1][:, :2], 0, values=0, axis=1
```
LucaMarconato, line 292: what does the -1 do here? Why is it needed?

src/napari_spatialdata/_view.py
```diff
@@ -256,12 +270,57 @@ def _select_layer(self) -> None:
...
+            )
+
+        self.model.spot_diameter = np.array([0.0, 10.0, 10.0])
```
LucaMarconato, line 295: I quickly checked the codebase and I believe that `spot_diameter` is never used. If so, I would remove it altogether.

src/napari_spatialdata/_view.py
```diff
@@ -256,12 +270,57 @@ def _select_layer(self) -> None:
...
+        self.model._labels_key = layer.metadata["region_key"] if isinstance(layer, Labels) else None
```
LucaMarconato, line 296: Ok for now, but we should probably get rid of the legacy `label_key` in a follow-up PR.

src/napari_spatialdata/_view.py
```diff
@@ -256,12 +270,57 @@ def _select_layer(self) -> None:
...
+            layer.metadata["adata"] = adata
+
+        if layer is not None and "adata" in layer.metadata:
+            with self.model.events.adata.blocker():
+                self.model.adata = layer.metadata["adata"]
+
+        if self.model.adata.shape == (0, 0):
+            return
+
+        if "spatial" in self.model.adata.obsm:
```
LucaMarconato, line 290: Ok for now, but we will have, as discussed, get rid of the legacy use of `AnnData`. An issue is already tracking this: https://github.com/scverse/napari-spatialdata/issues/180

src/napari_spatialdata/_view.py
LucaMarconato: Tracked here: https://github.com/scverse/napari-spatialdata/issues/180

src/napari_spatialdata/_view.py
LucaMarconato: Tracked here: https://github.com/scverse/napari-spatialdata/issues/180

src/napari_spatialdata/_view.py
LucaMarconato: I believe that this is not addressed yet: https://github.com/scverse/napari-spatialdata/issues/181. I believe that this could be quickly fixed by adding a call to `join_spatialelement_table` in the current function.

PR #199, opened 2024-03-04
Multi table (#198)

melonora:
implement multi_table in napari-spatialdata.


src/napari_spatialdata/_model.py
```diff
@@ -19,19 +18,15 @@ class ImageModel:
     """Model which holds the data for interactive visualization."""
```
LucaMarconato, line 18: A bit off-topic for this PR, but maybe I'd rename the `ImageModel` class to `DataModel`

src/napari_spatialdata/_view.py
```diff
@@ -98,31 +106,83 @@ def export(self) -> None:
...
+    def _update_adata(self) -> None:
+        if (table_name := self.table_name_widget.currentText()) == "":
+            return
+        layer = self._viewer.layers.selection.active
+        adata = None
+
+        if sdata := layer.metadata.get("sdata"):
+            element_name = layer.metadata.get("name")
+            table = sdata[table_name]
+            adata = table[table.obs[table.uns["spatialdata_attrs"]["region_key"]] == element_name]
```
LucaMarconato, line 118: This is an additional place where the join operation should be put.

src/napari_spatialdata/_view.py
```diff
@@ -256,12 +326,63 @@ def _select_layer(self) -> None:
...
+        else:
+            return
+
     def _get_adata_layer(self) -> Sequence[Optional[str]]:
         adata_layers = list(self.model.adata.layers.keys())
         if len(adata_layers):
             return adata_layers
         return [None]
 
+    def _get_init_table_list(self) -> Optional[Sequence[Optional[str]]]:
```
LucaMarconato, line 374: also this function is already present above (they are identical up to an empty line present here and not above)

src/napari_spatialdata/_viewer.py
```diff
@@ -403,24 +424,47 @@ def add_sdata_points(self, sdata: SpatialData, key: str, selected_cs: str, multi
             gen = np.random.default_rng()
             subsample = np.sort(gen.choice(len(points), size=100000, replace=False))  # same as indices
 
+        # TODO consider subsampling adata and passing that on.
+        adata, table_name, table_names = self._get_table_data(sdata, original_name)
```
LucaMarconato, line 428: Here the function `_get_table_data` returns `adata`, but this is not passed to `add_points`. This could be a double problem:
1. the adata (and the eventual join) performed in `_get_table_data()`, could be expensive, and they are not used
2. the subset operation that is done below in-line in `add_points()` is not a full join, so in the case in which the table mismatches the points it would lead to errors.

I would address this here by replacing the call of `_get_table_data(sdata)` with `get_table_data(...)` where the argument is a subset (=copied) of the points. If this is done I think this will address the TODO, so it can be removed.

src/napari_spatialdata/_model.py
melonora: lol, this is indeed what I call it in another plugin xD

src/napari_spatialdata/_view.py
```diff
@@ -256,12 +326,63 @@ def _select_layer(self) -> None:
         else:
             return
 
+    def _update_adata(self) -> None:
+        if (table_name := self.table_name_widget.currentText()) == "":
```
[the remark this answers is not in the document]
melonora: This is for different widgets. The `ScatterWidget` and `AdataViewWidget` do not share the `ImageModel`.

PR #201, opened 2024-03-06
Avoid exception when registering shortcut

aeisenbarth:
Napari does not allow registering an existing shortcut.

This can occur when:
- Another plugin or Napari itself or the user has registered "Shift-L", "Shift-E" before opening napari-spatialdata
- The user opened and closes napari-spatialdata multiple times in the same Napari session

As a consequence, Napari raises an exception and all code after shortcut registration cannot be used anymore.

This PR adds the optional argument `override` to silence the error. If you prefer `override=True`, I can change that (should have the same effect on avoiding the error).


LucaMarconato:
Thanks @aeisenbarth, merging it. Please notice that the `labels_key` usage will be removed when the code from https://github.com/scverse/napari-spatialdata/issues/180, which is currently merged to the `multi_table` branch, will be merged to `main`.

PR #205, opened 2024-03-13
Issue #180

### LINKED ISSUE #180: Clean up legacy code in `_models.py` and related parts

melonora:
closes #180 

This PR depends on #204 and the `multi_table` branch of `SpatialData` and addresses all the tasks in #180 as far as they were still applicable. One difference is that it both removes `spot_diameter` and `point_diameter` as they didn't seem to have an effect. Ultimately `spot_diameter` was used in case someone clicked on an item of `AListWidget` which for some reason would add a new napari layer if the ndim of the vector was equal to 2. This has been removed.

### DIFF
--- a/src/napari_spatialdata/_model.py
+++ b/src/napari_spatialdata/_model.py
@@ -27,7 +27,6 @@ class ImageModel:
     _color_by: str = field(default="", repr=True, init=False)
     _system_name: Optional[str] = field(default=None, repr=True)
 
-    _spot_diameter: Union[NDArrayA, float] = field(init=False, default=1)
     _scale_key: Optional[str] = field(init=False, default="tissue_hires_scalef")  # TODO(giovp): use constants for these
 
     _palette: Optional[str] = field(init=False, default=None, repr=False)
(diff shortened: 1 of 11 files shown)


LucaMarconato:
Looks great to me thanks! 🔥

LucaMarconato (approved):

PR #208, opened 2024-03-13
Test spatialelement table join

LucaMarconato:
Closes https://github.com/scverse/napari-spatialdata/issues/181

Adds a test for checking that the plot is correct even when some rows of the table are not annotating instances of the spatial element or when the spatial element contains instances not in the table.

Also, checks that this works both for the case when the `instance_key` column has dtype `int` and when it's a `str`.

Eventually, the `instance_key` column will only have dtype `int`, as discussed here: https://github.com/scverse/spatialdata/issues/158, but for the moment it would be easier if both were supported, as this PR will check.


src/napari_spatialdata/_model.py
```diff
@@ -78,7 +79,12 @@ def get_obs(
         """
         if name not in self.adata.obs.columns:
             raise KeyError(f"Key `{name}` not found in `adata.obs`.")
-        return self.adata.obs[name], self._format_key(name)
+        if name != self.instance_key:
+            adata_obs = self.adata.obs[[self.instance_key, name]]
+            adata_obs.set_index(self.instance_key, inplace=True)
+        else:
+            adata_obs = self.adata.obs
+        return adata_obs[name], self._format_key(name)
```
LucaMarconato, line 87: I find this function a bit confusing. Here is why:
- In general, I would use the `get_values()` function from `SpatialData`. But here (at leasta for the time being) since we are using `AnnData` as a data model and since there is the subset operation that complicates things, let's keep it like this.

Anywhay:
- [ ] I find confusing that the behavior of the function is different depending on the value being passed being instance_key or not. Why does this happen? I woudl use two seprate functions.
- [ ] Also, the instance_key columns may be numeric (most of the time it is), while `AnnData` converts numerical columns to string. This may create problems, please consider this in this case.

LucaMarconato: `get_obsm()` is actually not supported by `get_values()`, so that's another argument for keeping things like this for the time being.

src/napari_spatialdata/_viewer.py
```diff
@@ -417,35 +420,39 @@ def add_sdata_points(self, sdata: SpatialData, key: str, selected_cs: str, multi
...
-            subsample = np.arange(len(points))
+            subsample = None
         else:
             logger.info("Subsampling points because the number of points exceeds the currently supported 100 000.")
             gen = np.random.default_rng()
-            subsample = np.sort(gen.choice(len(points), size=100000, replace=False))  # same as indices
-
-        # TODO consider subsampling adata and passing that on.
-        adata, table_name, table_names = self._get_table_data(sdata, original_name)
+            subsample = np.sort(gen.choice(len(points), size=POINT_THRESHOLD, replace=False))  # same as indices
```
LucaMarconato, line 429: excellent

tests/test_spatialdata.py
```diff
@@ -1,109 +1,118 @@
...
+            blobs_extra_cs["table"].obs[blobs_extra_cs["table"].uns["spatialdata_attrs"]["region_key"]]
+            == list(blobs_extra_cs.labels.keys())[0]
         ).sum()
     )
     assert widget.viewer_model.viewer.layers[0].metadata.get("region_key") is not None
 
 
-def test_sdatawidget_points(caplog, make_napari_viewer: Any):
-    sdata.points["many_points"] = PointsModel.parse(
+@pytest.mark.xfail
```
LucaMarconato, line 93: I would remove this and adjust the asserts in the test.

src/napari_spatialdata/_model.py
melonora: yeah the latter comment is exactly why I did not go for that approach. As for the different behaviour, in the `else` part, the user is basically wanting to plot the name of the `instance_key` column. If we would use the code in the `if` conditional this would result in a duplicate column which then results in a problem when setting the index.

melonora: Btw regarding your second point, this does not get converted to string as we are not returning a full anndata object here. We are working with a subset of obs so it is not validated by anndata and thus not converted.

LucaMarconato:
Some additional comments, I will put them here while I test.
- [ ] Points from the `merfish` dataset can't be visualized.

LucaMarconato:
- [ ] Unfortunately also genes from the `mibitof` dataset can't be visualized (I checked them because they contain labels/table mismatches).
```
File ~/embl/projects/basel/napari-spatialdata/src/napari_spatialdata/_widgets.py:217, in AListWidget._(self=<napari_spatialdata._widgets.AListWidget object>, vec=cell_id
2      -0.042887
4      -0.004494
5     ...-0.039209
Name: vec, Length: 1023, dtype: float32, **kwargs={'key': 'CD31'})
    215 vec = pd.Series(vec, name="vec", index=instance_key_col)
    216 layer_meta = self.model.layer.metadata if self.model.layer is not None else None
--> 217 element_indices = pd.Series(layer_meta["indices"], name="element_indices")
        layer_meta = {'sdata': SpatialData object with:
├── Images
│     ├── 'point8_image': SpatialImage[cyx] (3, 1024, 1024)
│     ├── 'point16_image': SpatialImage[cyx] (3, 1024, 1024)
│     └── 'point23_image': SpatialImage[cyx] (3, 1024, 1024)
├── Labels
│     ├── 'point8_labels': SpatialImage[yx] (1024, 1024)
│     ├── 'point16_labels': SpatialImage[yx] (1024, 1024)
│     └── 'point23_labels': SpatialImage[yx] (1024, 1024)
└── Tables
      └── 'table': AnnData (3309, 36)
with coordinate systems:
▸ 'point8', with elements:
        point8_image (Images), point8_labels (Labels)
▸ 'point16', with elements:
        point16_image (Images), point16_labels (Labels)
▸ 'point23', with elements:
        point23_image (Images), point23_labels (Labels), 'adata': AnnData object with n_obs × n_vars = 1023 × 36
    obs: 'row_num', 'point', 'cell_id', 'X1', 'center_rowcoord', 'center_colcoord', 'cell_size', 'category', 'donor', 'Cluster', 'batch', 'library_id'
    uns: 'spatialdata_attrs'
    obsm: 'X_scanorama', 'X_umap', 'spatial', 'region_key': 'library_id', 'instance_key': 'cell_id', 'table_names': ['table'], 'name': 'point16_labels', '_active_in_cs': {'point16'}, '_current_cs': 'point16'}
        pd = <module 'pandas' from '/Users/macbook/miniconda3/envs/ome/lib/python3.10/site-packages/pandas/__init__.py'>
    218 diff_element_table = set(instance_key_col).symmetric_difference(element_indices)
    219 merge_vec = pd.merge(element_indices, vec, left_on="element_indices", right_index=True, how="left")[
    220     "vec"
    221 ].fillna(0, axis=0)
    ```

LucaMarconato:
I could reproduce the bug appearing with the `mibitof` dataset in `test_partial_table_matching_with_arbitrary_ordering()` (it appears both for labels and multiscale_labels`. I could not reproduce the points bug with the test (because no subsampling is happening).

I verified the robustness against shuffling and it works. I will now check the `get_extent()` code.

PR #210, opened 2024-03-19
Napari annotation

melonora:
annotation widget for napari-spatialdata


LucaMarconato:
As discussed with @melonora, here is a list of the following tasks to finalize the annotation widget:
- [ ] the annotations saved with the annotation widget are saved in the `GeoDataFrame` columns. These currently cannot be visualized by napari (we need to replicate the code used for the `DaskDataFrame` columns).
- [ ] we need an extra widget for visualizing the obs of the table or the shapes/points dataframe. This widget should enable the modification of the dataframe column so that the user can add text annotation (like the specific name of an annotated region), or the name of the annotator itself

melonora:
I will create a barebone PR at the moment and implement the table widget afterwards.

LucaMarconato:
After discussion with @melonora, better to start by merging https://github.com/scverse/napari-spatialdata/pull/233 and then proceed by copying over the part on the annotation table from this PR.

PR #224, opened 2024-03-26
warn many shapes as points

### LINKED ISSUE #213: Warn users that layers with many shapes are slow to render
A limitation of napari is the ability to visualize large collections (>50k) shapes due to performance limitations of the current rendering approach. This can lead to a undesirable user experience. We should add a messagebox warning the users about this when such a layer is requested to be shown.

Side note, we are exploring workarounds for this, like on-the-fly rasterization of the shapes into a multiscale, chunked labels image.

melonora:
closes #213 

This PR checks whether there is a point geometry in a shapes element and if the length of the element exceeds 1m adds an icon to the list widget item. When a user then hovers over the item they get to see a warning regarding visualization.


melonora:
This could also potentially be extended to put warning signs and tooltips for the subsampling of points.

LucaMarconato:
Thanks, this looks great! As discussed I have reduced the threshold, extended this to polygons/multipolygons (with a lower threshold) and moved the constants to the `config.py` file.

PR #226, opened 2024-03-28
Avoid exception when key binding collides with existing

aeisenbarth:
This avoids raising an error on key binding collision when napari-spatialdata is opened/closed multiple times (or the keymap has bin registered by another plugin).

Sorry, my previous fix was in the wrong direction and too permissive. There is no way to "register only if not yet in use", but one can only force override any previous shortcut. Without `override=True`, the error is raised.


LucaMarconato:
Thanks a lot!

PR #233, opened 2024-04-09
Barebone annotate

melonora:
This PR serves to provide a barebone annotation widget. At the moment only shape annotations are supported but I am to extend this to points and perhaps labels as well. Furthermore, at a later stage also a table widget will be provided.

Currently, the user can annotate using and also reload shape annotations. The `description` column at the moment serves no purpose but this will come with the table widget.

@LucaMarconato maybe before putting this in we need to discuss a little bit about UI experience.


melonora:
a class is added but the class is not annotating any element. In this case, I suggest to just ignore the class when saving to disk, this is already there. I checked again with the table widget btw, single molecule didn't show anything because it is the cell element that the table is annotating. Checked with cell and it properly shows.

melonora:
Regarding an element is not belonging to any class; saving should still work.

This works by setting the class to undefined.

LucaMarconato:
Small comment:
- [ ] please document the code inside `on_mouse_move()` in `_view.py`.

LucaMarconato:
I tested the saving (on the incremental io branch). I found some bugs:
- [ ] when saving an element, the View widget is not updated. Workaround: from the element widget add the element, and delete it; now the View widget is updated an it works.
- [ ] The colors of the annotation widget do not correspond to the colors used for plotting, easy fix.
- [ ] The "Annotation table" dropdown combobox does not update the color classes.

Saving and overwriting work!

LucaMarconato:
Bug that I probably introduced here (I was working on performance in order to avoid copies) https://github.com/scverse/napari-spatialdata/pull/239/files in `_model.py`:
- [ ] when trying to plot `instance_key` an error is raised. The user playing around with the annotation widget may encounter this bug because the table created has only a few columns, so it's likely to try to plot `instance_id`.

melonora:
> I tested the saving (on the incremental io branch). I found some bugs:
> 
> * [x]  when saving an element, the View widget is not updated. Workaround: from the element widget add the element, and delete it; now the View widget is updated an it works.
> * [x]  The colors of the annotation widget do not correspond to the colors used for plotting, easy fix.
> * [ ]  The "Annotation table" dropdown combobox does not update the color classes.
> 
> Saving and overwriting work!

not certain what you mean with the last point

melonora:
tests pass locally, but here it would require a release of spatialdata with the incremental IO. For hackathon purposes merging now already.

LucaMarconato:
>  The "Annotation table" dropdown combobox does not update the color classes.

I will explain this in a new issue, I have just verified and it is still open.

PR #239, opened 2024-05-07
Labels support for 0.4.19, points support for > 0.4.19

LucaMarconato:
- labels require `colormap` instead of `color` from `0.5.0`
- points require `border_xxx` instead of `edge_xxx` from the latest `main` (unreleased, > `0.5.0`).

This PR assumes that the next released version of napari is `0.5.0` (the current is `0.4.19`). @melonora confirmed that this should be the case. If not, we need to adjust the version strings in this PR.

Also, with @melonora we discussed if napari should automatically accept both the old argument and the new argument, as we do with our deprecation decorator approach: https://github.com/scverse/spatialdata/pull/562. If it will, then we can remove the manual versions if-else checks added with this PR.

### DIFF
--- a/src/napari_spatialdata/_model.py
+++ b/src/napari_spatialdata/_model.py
@@ -85,12 +85,12 @@ class DataModel:
         if name not in self.adata.obs.columns:
             raise KeyError(f"Key `{name}` not found in `adata.obs`.")
         if name != self.instance_key:
-            adata_obs = self.adata.obs[[self.instance_key, name]].copy()
-            adata_obs.set_index(self.instance_key, inplace=True)
+            obs_column = self.adata.obs[[self.instance_key, name]]
+            obs_column = obs_column.set_index(self.instance_key)[name]
         else:
-            adata_obs = self.adata.obs.copy()
-            adata_obs.index = adata_obs[self.instance_key]
-        return adata_obs[name], self._format_key(name)
+            obs_column = self.adata.obs[name].copy()
+            obs_column.index = self.adata.obs[self.instance_key]
+        return obs_column, self._format_key(name)
 
     @_ensure_dense_vector
     def get_columns_df(self, name: Union[str, int], **_: Any) -> Tuple[Optional[NDArrayA], str]:
--- a/src/napari_spatialdata/_model.py
+++ b/src/napari_spatialdata/_model.py
@@ -85,10 +85,11 @@ class DataModel:
         if name not in self.adata.obs.columns:
             raise KeyError(f"Key `{name}` not found in `adata.obs`.")
         if name != self.instance_key:
-            adata_obs = self.adata.obs[[self.instance_key, name]]
+            adata_obs = self.adata.obs[[self.instance_key, name]].copy()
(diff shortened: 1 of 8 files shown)


LucaMarconato:
Tests pass locally.

src/napari_spatialdata/_model.py
```diff
@@ -85,10 +85,11 @@ def get_obs(
         if name not in self.adata.obs.columns:
             raise KeyError(f"Key `{name}` not found in `adata.obs`.")
         if name != self.instance_key:
-            adata_obs = self.adata.obs[[self.instance_key, name]]
+            adata_obs = self.adata.obs[[self.instance_key, name]].copy()
```
melonora, line 88: is this copy required because of the otherwise automated conversion to categorical when plotting?

src/napari_spatialdata/_viewer.py
```diff
@@ -539,12 +548,13 @@ def add_sdata_points(self, sdata: SpatialData, key: str, selected_cs: str, multi
         np.fliplr(xy)
         # radii_size = _calc_default_radii(self.viewer, sdata, selected_cs)
         radii_size = 3
+        version = get_napari_version()
+        kwargs = {"edge_width": 0.0} if version <= packaging.version.parse("0.4.19") else {"border_width": 0.0}
```
melonora, line 552: I would say this solution is very temporary. 0.4.19 had some fixes with layer.events.data that will be important for the table widget so we would need to have 0.4.19 as constraint.

melonora (approved):
Pre-emptive approval. I will check regarding the leaking problem. Might have to close the viewer prior to the test that is failing. 

If you could add a todo perhaps to change the constraint to 0.4.19 or higher as soon as the table widget is in. In 0.4.18 there were quite some issues with `layer.events.data`. In 0.4.19 there were still some but way less and we require these events for a properly working table widget.

src/napari_spatialdata/_model.py
LucaMarconato: Good point, I could remove it. It was because I was setting the index in-place.

src/napari_spatialdata/_widgets.py
```diff
@@ -136,8 +132,14 @@ def _onAction(self, items: Iterable[str]) -> None:
                     self.model.layer.face_color = properties["face_color"]
                     self.model.layer.text = properties["text"]
                 elif isinstance(self.model.layer, Labels):
-                    self.model.layer.color = properties["color"]
-                    self.model.layer.properties = properties.get("properties", None)
+                    version = get_napari_version()
+                    if version < packaging.version.parse("0.4.19"):
```
melonora, line 136:
```suggestion
                    if version < packaging.version.parse("0.5.0"):
```

LucaMarconato:
Good to merge now. With @melonora we discussed if napari should automatically accept both the old argument and the new argument, as we do with our deprecation decorator approach: https://github.com/scverse/spatialdata/pull/562. If it will, then we can remove the manual versions if-else checks added with this PR.

PR #254, opened 2024-06-12
showing channel names of images in var list widget

LucaMarconato:
The feature is already in main but tests do not pass (partially because a new release in `spatialdata` is required. Let's fix this in this PR after a release is available.


LucaMarconato:
Actually, I am going to merge this as it is and we can fix the tests in a follow up PR.

Issue #260, opened 2024-06-24
Problems when saving multiple annotations tables for the same element


LucaMarconato:
I have observed problems when saving multiple annotation tables for the same shapes element using the annotation widget. Not sure if it's a macOS specific bug @melonora. Here is how to reproduce:

- open `napari-spatialdata` and a layer from `SpatialData` object
- create a new shapes layer
- open the annotation widget
- link the layer to the sdata
- create some classes
- create some annotations around the classes
- save the annotation table

So far all works great.

- add a new annotation class, create a new annotation
- [x] save the annotation using the same element name but a new annotation table. This leads to a warning (something like `Shapes_external and annotation_Shapes_external2 already exist. Do you want to overwrite them?`). The message is incorrect. It should have an "or" instead of the "and".
- [x] If we save (but also if we don't save I think), the combobox "Annotation table:" will add some duplicate entries.
- [x] If I added extra shapes before saving the second time, then going through the entries of the "Annotation table:" combobox would lead to warnings like `UserWarning: The provided face_color parameter has 4 entries, while the data contains 5 entries. Setting face_color to white.!` and to the annotations being displayed as white for some of the entries instead of with the class colors.

Finally:
- [ ] if I save the annotations under a new element and under a new table name, the original shapes layer gets renamed to the new name. Therefore, since we are not really saving a new element but updating the original one, I would consider the following:
	- ~~when "Save selected annotation layer" is pressed, the user can only choose the name of the annotation table~~ edit: we opted for letting the user choose both names
	- [ ] and the shapes layer is always overwritten in place.

melonora:
will check

LucaMarconato:
> and the shapes layer is always overwritten in place.

can you confirm that now with #261 the shapes layer is overwritten in place? At least I see that the old layer disappears from the view and is replaced by the new layer (with eventually a new name). I guess therefore that also on disk the old layer (and old name) disappears. If so, we can tick this last checkbox.

melonora:
see description of #261

melonora:
TLDR yes it is. The old shape will still exist though when the user changed both the table name and the shapes name.

PR #261, opened 2024-06-24
Fix annotation

melonora:
This PR solves the issues mentioned in #260.  The save message has been corrected and there are also no duplicate table names.
Regarding the latter point of adding new entries, saving and then going back to the old table leading to issues with non matching indices, I opted to delete the previous table. The other option would be to do a left exclusive join to see if there are indices not present in the table and then specifically add a row, but this seemed a bit more cumbersome. However this can be discussed.

Regarding the final point, I think it is valid that if a user changes the name, the user is adding an element. This allows the user to still diverge from a given table by changing both the element name and the table name.


LucaMarconato (approved):

melonora:
The issue you mention seems to be mac specific, I don't have this at all. Then again with the napari 0.5.0 alpha release that just went down we might have to check this.

Issue #262, opened 2024-06-25
Bug with contrast limit calculation for multiscale images


LucaMarconato:
The code changes added here regarding the contrast limit adjustment: https://github.com/scverse/napari-spatialdata/commit/d639dd7e24266305d0f79484b32bea2f3902d6f4
lead to an error when changing channel for a multiscale multichannel image. This can be reproduced with `morphology_focus` from this Xenium Multimodal Segmentation dataset: https://github.com/giovp/spatialdata-sandbox/tree/main/xenium_2.0.0_io.

I discussed this with @melonora, we would need to:
- [x] fix the how the contrast limit is computed in `_view.py`
- [x] preferably add a comment in `_view.py` in which `has_anndata` is used (because an AnnData object that shows the channel names on the View widget is added).

PR #264, opened 2024-06-26
simplify conditional

melonora:
some commits are in napari-spatialdata repo already, but somehow are not recognized by github on my fork


LucaMarconato:
Mmm weird. Would it be the fix for https://github.com/scverse/napari-spatialdata/issues/262?

melonora:
nope, I opened a PR that will fix #262. I don't know what is going on with that fork though as you see the exact same commits in the PR that I opened as fix for it.

PR #265, opened 2024-06-26
fix contrast limits

### LINKED ISSUE #262: Bug with contrast limit calculation for multiscale images

melonora:
closes #262 

Fixes the issue in #262 regarding calculating contast limits for MultiScale data. This PR checks whether data is multiscale and if so calculates the contrast limits based on the lowest resolution. Furthermore, an additional check is added to see whether contrast limits are monotonically increasing and if this is not the case it sets the contrast limits to the dtype range.


LucaMarconato:
LGTM!

LucaMarconato (approved):

PR #270, opened 2024-07-11
pyqt scatter widget

fjorka:
### Description
This PR introduces a complete Scatter Widget based on PyQtGraph, intended to replace the Matplotlib-based Scatter Widget.

### Rationale
- Expected improved performance for large datasets.
- Smooth navigation for zooming and panning within the plot.
- Increased flexibility in customization.

### Functionality
- It preserves all functionality of the previous widget.
- Documentation notebooks updated with the information and screenshots of the new widget.
- New functionality includes:
    - Separate color controls for continuous and discrete data:
        - Continuous color control offers:
            - Histogram of values chosen for color.
            - Ability to change LUT.
            - Control over contrast.
        - Discrete color control offers:
            - Scrollable legend of colors.
            - Ability to change the color for a selected class.
    - Pseudohistogram plot when only one axis is defined.
    - Highlight of nearby data points under cursor.
    - Dynamic status display of cursor position and highlighted data points.
    - Expanded ROI annotations:
         - Lasso or rectangle tool (useful for thresholding).
         - Ability to use several disjointed ROIs for annotation.
         - Modifiable ROIs.
     - Dialog to choose Annotation name (previously Export).
     - Saving options:
         - When started with AnnData, the Save button opens a path dialog to choose a location for an AnnData file (supports h5ad, zarr, and csv).
         - When opened with the Napari viewer, it mimics the behavior of the Annotation Widget:
             - With backed sdata, Save will modify sdata of the selected layer.
             - Without backed sdata, no saving action.
         - The widget displays the status and saving possibility when started. 

### Next Steps Beyond This PR
 - [ ] Flip the Y-axis (scatterplot in spatial coordinates is flipped relative to the image).
 - [ ] Signal point highlight to the main viewer for quick connection between visualizations in different dimensions.
 - [ ] Highlight object in the main Napari window to trigger highlight in the scatterplot.
 - [ ] Integrate Annotation Widget functionality into the Scatter Widget (the Annotation Widget facilitates annotating multiple classes within a single column).
 - [ ] Performance improvements (1. How points belonging to polygons is checked.)


src/napari_spatialdata/_view.py
```diff
@@ -32,14 +34,44 @@
...
+    ComponentWidget,
+    RangeSliderWidget,
+)
 
 __all__ = ["QtAdataViewWidget", "QtAdataScatterWidget"]
 
 from napari_spatialdata.utils._utils import _get_init_table_list, block_signals
 
 
+class AnnotationDialog(QDialog):
```
melonora, line 50:
```suggestion
class ScatterAnnotationDialog(QDialog):
```
Perhaps this is better to avoid ambiguity regarding what annotation is meant as we now also have the annotation widget:)

melonora:
oh I see a green status! Ready for review?

fjorka:
> oh I see a green status! Ready for review?

Not yet, almost :)

melonora:
Thanks @fjorka, I was trying it out on a windows machine. For now I seem to have caught a bug when loading the widget and then moving the mouse over it: ```AttributeError: 'QPointF' object has no attribute 'localPos'```, happening with the call of `super().mouseMoveEvent`. Will look into it.

LucaMarconato:
Trying this now 😊 @melonora I get the same error on macOS.

LucaMarconato:
Here is a first initial round of review. I start with a review of the interface by trying it out (I didn't check the code yet), and of the documentation notebooks.
- In my latest commit I propose a fix for the bug affecting me and Wouter. @melonora can you double check it?
- in my macOS interface I don't see the lasso and rectangle buttons 🥸 any idea what's going on? @melonora do you see them? I can dig more into this.
- The notebooks look great! I would only make one thing clearer in the "scatterwidget.ipynb" and "scatterwidget_annotation.ipynb" notebooks. 
I would clarify a bit around the scenarios: "napari is running" vs "no napari viewer detected" and the case in which the table is from a standalone `AnnData` vs coming from a `SpatialData` object. 
In principle there are 4 cases. Are these two possible: "napari is running" and "the table is from a standalone `AnnData`" and "no napari viewer detected" and "the table is from a `SpatialData` object?
- is it possible to remove the white border around the scatterplot points?

I'll proceed next with the review of the code 😊 Looking forward to have this in main, super cool work!

setup.cfg
```diff
@@ -87,9 +88,11 @@ doc =
     myst-nb
     # used in the notebooks
     squidpy
+    pyqtgraph
 
 readthedocs =
 	pydantic<2
+    pyqtgraph
```
LucaMarconato, line 95: wrong indentation?

LucaMarconato:
Segfault in the tests 👀

LucaMarconato:
I re-run the tests, now we get a proper error (not a segfault):

```
  tests/test_scatterwidgets.py::test_add_roi_to_plot PASSED                [ 31%]
  tests/test_scatterwidgets.py::test_remove_roi_double_click PASSED        [ 32%]
  tests/test_scatterwidgets.py::test_remove_roi_double_click ERROR         [ 32%]
  tests/test_scatterwidgets.py::test_remove_hovered_roi PASSED             [ 33%]
```

src/napari_spatialdata/_view.py
```diff
@@ -115,12 +170,167 @@ def __init__(self, napari_viewer: Viewer, model: DataModel | None = None):
...
+                    ]
+
+                    new_table = pd.merge(
+                        sel_obs, self.model.adata.obs[merge_on + columns_to_add], on=merge_on, how="left"
+                    )
+                    new_table[self.annotation_name] = new_table[self.annotation_name].astype("category")
+                    selected_layer.metadata["sdata"][selected_table].obs = new_table
+
+                    # trigger update of the viewer
+                    list(self._viewer.window._dock_widgets.items())[-2][1].children()[4]._on_layer_update()
```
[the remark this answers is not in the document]
LucaMarconato: I would also replace the `[1]`, if possible, with something more explicit to avoid potential bugs like the one above. If this is not possible, I would add a comment explaining what `[1]` will access.

LucaMarconato:
Ok I finished testing the interface, to be organized I suggest to keep action items clearly discoverable by proceeding as follow:
- tasks are either in the form of conversations; until `Resolve conversation` is pressed, the task is open
- [x] either in the form of check box items, like this one
- either posted automatically as comments from ReviewNB, in that case we can consider the task open until a 👍 is given.

Also, I suggest to try to self-assign/split some of the tasks. I'll list mine below:
- [x] (Luca) test of the interface (didn't look at the code yet)
- [x] (Luca) reviewed the notebooks
- [ ] (Luca) reproduce locally with Python 3.9 and try to understand why the CI fails
- [ ] (Luca) try to fix the problem with the lasso and rectangle selection button that appear as black in macOS.
- [ ] (Luca) review the code

Czaki:
In context of failing test. 
The `widget` created in the test is not either added to `qtbot` using `qtbot.add_widget` or docked into napari viewer so may be cleaned before all events were proceeded, that leads to segfault. 

It may be impossible to reproduce on a local machine.

Czaki:
You may want to read this discussion until we improve the documentation https://napari.zulipchat.com/#narrow/stream/212875-general/topic/qtbot.20documentation/near/462664015

LucaMarconato:
Thanks @Czaki for the explanation, I indeed can't reproduce the failing test on my machine with Python 3.9, and running the test again showed a segfault for a different test.

LucaMarconato:
I checked, the `test_remove_roi_double_click()` ([code here](https://github.com/fjorka/napari-spatialdata/blob/kmk-scatter-pg/tests/test_scatterwidgets.py#L278)), which was giving me a segfault, seems to handle the `qtbot` correctly, [as seen here](https://github.com/fjorka/napari-spatialdata/blob/fb2f513e98971a28a2ec5d89c20c1590d56c0854/tests/test_scatterwidgets.py#L31).

Similarly, the test `test_scatterlistwidget()` ([code here](https://github.com/scverse/napari-spatialdata/blob/65d27a7e921960851f4bb9646076dcb401c5862e/tests/test_widgets.py#L120C5-L120C27)), which now is giving a segfault, seems to be calling `make_napari_viewer()` correctly ([code here](https://github.com/fjorka/napari-spatialdata/blob/fb2f513e98971a28a2ec5d89c20c1590d56c0854/tests/test_widgets.py#L131)).

I'll move away from this issue for the moment and start reviewing the code and get back to this later. Meanwhile if anyone wants to take a look at this, you are welcome!

tests/test_widgets.py
```diff
@@ -256,3 +264,56 @@ def test_layer_selection(make_napari_viewer: Any, image: NDArrayA, widget: Any,
...
+    mocker.patch.object(scatter_widget.plot_widget, "get_selection", return_value=annotation_values)
+
+    scatter_widget.export()
+
+    assert scatter_widget.status_label.text() == "Status: Annotation added."
+    assert np.array_equal(scatter_widget._model.adata.obs.test, annotation_values)
+
+
+def test_old_annotation(adata_labels, annotation_values, mocker):
+    """Test export - adding a new annotation."""
```
LucaMarconato, line 307: Please correct the docstring. Here the docstring is with `new`, but it should refer to an "old" annotation (as in the function name). Maybe explain that with old it means an alredy existing annotation that is updated.

tests/conftest.py
```diff
@@ -118,6 +134,43 @@ def labels():
     return blobs
 
 
+@pytest.fixture
+def prepare_cont_test_data():
```
LucaMarconato, line 138: I'd use the full name for improving readability, so `continuous` in place of `cont`.

tests/conftest.py
```diff
@@ -118,6 +134,43 @@ def labels():
...
+    color_data = {"vec": color_vec}
+
+    x_label = generate_random_string(10)
+    y_label = generate_random_string(10)
+    color_label = generate_random_string(10)
+    return x_data, y_data, color_data, x_label, y_label, color_label
+
+
+@pytest.fixture
+def prepare_disc_test_data():
```
LucaMarconato, line 154: Same here, I'd use `discrete`

tests/test_scatterwidgets.py
```diff
@@ -1,58 +1,404 @@
...
+    plot_widget._onClick(*prepare_cont_test_data)
+    plot_widget.plot()
+
+    assert plot_widget.x_data is not None
+    assert plot_widget.y_data is not None
+    assert len(plot_widget.scatter.xData) == DATA_LEN
+    assert len(plot_widget.scatter.yData) == DATA_LEN
+
+
+def test_plot_data_cont(plot_widget, prepare_cont_test_data):
```
LucaMarconato, line 58: same here, I'd use `continuous` instead of `cont`.

tests/test_scatterwidgets.py
```diff
@@ -1,58 +1,404 @@
...
+    """Test building lut widget."""
+    plot_widget._onClick(*prepare_cont_test_data)
+
+    assert plot_widget.lut is not None
+    assert plot_widget.discrete_color_widget is None
+
+
+def test_plot_data_disc(plot_widget, prepare_disc_test_data):
```
LucaMarconato, line 66: And here `discrete` instead of `disc`.

tests/test_scatterwidgets.py
```diff
@@ -1,58 +1,404 @@
...
+    """Test building disc colors widget."""
```
LucaMarconato, line 67: Same here, `discrete` instead of `disc`.

tests/test_scatterwidgets.py
```diff
@@ -1,58 +1,404 @@
...
+    plot_widget._onClick(*prepare_disc_test_data)
+
+    assert plot_widget.scatter is not None
+    assert plot_widget.lut is None
+    assert plot_widget.discrete_color_widget is not None
```
LucaMarconato, line 72: Is it correct for this to be `not None` here?

tests/test_scatterwidgets.py
```diff
@@ -1,58 +1,404 @@
...
+
+
+def test_plot_data_widget_change(plot_widget, prepare_cont_test_data, prepare_disc_test_data):
```
LucaMarconato, line 75: The linter of VS Code says "Redefining name 'plot_widget' from outer scope (line 27) Pylint (W0621: redefined-outer-name)" in this line. If it is true that `plot_widget` is redefined and not using the function at line 27 (the one that calls `qtbot.addWidget(widget)`), this could explain why we get the segfault in the CI in the test (adding the widget to the qt bot should prevent the segfault, but if that code is not used, then the segfault is not prevented).

tests/test_scatterwidgets.py
```diff
@@ -1,58 +1,404 @@
...
+    plot_pos = pg.Point(x, y)
+    view_pos = plot_widget.scatter_plot.vb.mapViewToScene(plot_pos)
+    return plot_widget.scatter_plot.scene().views()[0].mapFromScene(view_pos)
+
+
+def send_mouse_event(event_type, widget, pos, button=QtCore.Qt.LeftButton):
+    """Send a mouse event to a widget."""
+    # necessary because of https://bugreports.qt.io/browse/QTBUG-5232
+    event = QtGui.QMouseEvent(event_type, pos, button, button, QtCore.Qt.NoModifier)
+    QtWidgets.QApplication.sendEvent(widget, event)
```
LucaMarconato, line 177: This workaround is for macOS (as the issue from the url in the comment explains), and it could explain why the segfault appear for macOS, as it could be that this mechanism is not being taken into account by the qtbot.

tests/test_scatterwidgets.py
LucaMarconato: (I actually don't think this is the case, but wanted to still share this)

tests/test_scatterwidgets.py
```diff
@@ -1,58 +1,404 @@
...
+    point1 = [10, 10]
+    point2 = [70, 70]
+    point3 = [10, 80]
+
+    # Convert plot coordinates to viewport coordinates
+    viewport_pos1 = plot_coords_to_viewport(plot_widget, *point1)
+    viewport_pos2 = plot_coords_to_viewport(plot_widget, *point2)
+    viewport_pos3 = plot_coords_to_viewport(plot_widget, *point3)
+
+    view_widget = plot_widget.scatter_plot.scene().views()[0].viewport()
```
LucaMarconato, line 197: what does `[0]` refers to? Could this cause problems?

tests/test_scatterwidgets.py
```diff
@@ -1,58 +1,404 @@
...
+    roi1 = pg.RectROI([0, 0], [0.5, 0.5])
+    roi2 = pg.RectROI([0.6, 0.6], [0.7, 0.7])
+    roi3 = pg.RectROI([0.8, 0.8], [0.9, 0.9])
+    plot_widget._onClick(*prepare_disc_test_data)
+    plot_widget.roi_list = [roi1, roi2, roi3]
+    plot_widget.plot()
+
+    assert len(plot_widget.roi_list) == 3
+
+    event = QtGui.QKeyEvent(QtCore.QEvent.KeyPress, QtCore.Qt.Key_E, QtCore.Qt.ShiftModifier)
```
LucaMarconato, line 342: please add a comment discusssing the testing of Shift+E.

tests/test_scatterwidgets.py
```diff
@@ -1,58 +1,404 @@
...
+    assert len(plot_widget.roi_list) == 3
+
+    roi2.mouseHovering = True
+    event = QtGui.QKeyEvent(QtCore.QEvent.KeyPress, QtCore.Qt.Key_D, QtCore.Qt.NoModifier)
+    QtWidgets.QApplication.sendEvent(plot_widget, event)
+    assert len(plot_widget.roi_list) == 2
+
+    event = QtGui.QKeyEvent(QtCore.QEvent.KeyPress, QtCore.Qt.Key_D, QtCore.Qt.ShiftModifier)
+    QtWidgets.QApplication.sendEvent(plot_widget, event)
+    assert len(plot_widget.roi_list) == 0
```
LucaMarconato, line 353: I'd also write a small one line comment for E and Shift+E telling that E removes the hoverered roi and Shift+E removes all the rois.

tests/test_scatterwidgets.py
```diff
@@ -1,58 +1,404 @@
...
+    plot_widget.roi_list = [roi]
+
+    boolean_vector = plot_widget.get_selection()
+
+    assert len(boolean_vector) == DATA_LEN
+    assert np.sum(boolean_vector) == DATA_LEN
+
 
-RNG = np.random.default_rng(seed=0)
+def test_auto_range_disc(plot_widget, prepare_disc_test_data):
```
LucaMarconato, line 368: in general I'd look for occurrences of "disc" and "cont" in the code and turn them into "discrete" and "continuous"

tests/test_scatterwidgets.py
```diff
@@ -1,58 +1,404 @@
...
+    """Test auto range for discrete data."""
```
LucaMarconato, line 369: I'd explain please what the test aims to test. That auto range doesn't affect custom data colors?

tests/test_scatterwidgets.py
```diff
@@ -1,58 +1,404 @@
...
+    assert np.allclose(plot_widget.discrete_color_widget.color_buttons[label].color().getRgbF(), org_color, atol=1e-03)
 
-def test_matplotlib_widget_plot(make_napari_viewer: Any):
-    viewer = make_napari_viewer()
-    x_data, y_data, color_data, x_label, y_label, color_label = prepare_test_data()
-    mpl_widget = MatplotlibWidget(viewer, DataModel)
 
-    mpl_widget._onClick(x_data, y_data, color_data, x_label, y_label, color_label)
+def test_auto_range_cont(plot_widget, prepare_cont_test_data):
+    """Test auto range for discrete data."""
```
LucaMarconato, line 388: Also here.

LucaMarconato (changes requested):

LucaMarconato (changes requested):
@fjorka I finished reviewing the code. The vast majority of the comments refer to minor points, with maybe the tricky part being only regarding the failing test on macOS with Python 3.9. In general, tring the UI shows that the widget works very well and the notebooks are great. Great work!

tests/test_scatterwidgets.py
LucaMarconato: CC @Czaki, what do you think?

tests/test_scatterwidgets.py
```diff
@@ -1,58 +1,404 @@
...
+
+
+@pytest.fixture
+def plot_widget(qtbot):
+    """Fixture for creating a PlotWidget instance."""
+    model = DataModel()
+    widget = PlotWidget(None, model)
+    qtbot.addWidget(widget)
+    yield widget
+    widget.deleteLater()
```
Czaki, line 33: Once you have added widget to qtbot, then you should not use `deleteLater`.

fjorka:
- [x] is it possible to remove the white border around the scatterplot points?

Replaced with black (invisible) border when color is defined - 1acdd85274de92eebc5cb8607113bf4152cef3cc

tests/test_scatterwidgets.py
fjorka: Hi @Czaki !

I do see the error:
```
E       AssertionError: Some instance of QtViewer is not properly cleaned in one of previous test. For easier debug one may use --save-leaked-object-graph flag for pytest to get graph of leaked objects. If you use qtbot (from pytest-qt) to clean Qt objects after test you may need to switch to manual clean using `deleteLater()` and `qtbot.wait(50)` later.
E       assert 1 == 0
```
when I run pytest locally. 

You are correct, the additional `deleteLater` is unnecessary. I added it following the error message as I was getting desperate. I will remove it to avoid confusion.

Now, the minimum example to reproduce the error is:
```
import pytest
from napari_spatialdata._model import DataModel
from napari_spatialdata._scatterwidgets import PlotWidget

@pytest.fixture
def plot_widget(qtbot):
    """Fixture for creating a PlotWidget instance."""
    model = DataModel()
    widget = PlotWidget(None, model)
    qtbot.addWidget(widget)
    yield widget

def test_initialization(plot_widget):
    """Test initialization of PlotWidget."""
    assert plot_widget is not None

def test_elementwidget(make_napari_viewer):
    _ = make_napari_viewer()
    assert 1 == 1
```

However, it's not the problem specific to this branch. If I remove all tests from test_scatterwidgets.py, the error persists:
image

Finally, the error is triggered only once. If I run the fake test (the code pasted in above) as test_aaa.py before any other test, the error appears in the fake test but is not triggered again by any of the original tests:
image

Any ideas what is going on here?

src/napari_spatialdata/_scatterwidgets.py
```diff
@@ -374,3 +202,846 @@ def clear(self) -> None:
...
+        self.layout.setSpacing(0)  # Set minimal spacing
+        self.setLayout(self.layout)
+
+        self.color_buttons = {}
+
+        # TODO- format of the palette in the model itself
+        # if self._model.palette is not None:
+        #     self.palette = self._model.palette
+        # else:
+        self.palette = label_colormap(len(color_data["labels"]))
```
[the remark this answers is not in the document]
fjorka: I left the 'TODO' annotation from the previous version so that we have it marked for the future.

tests/test_scatterwidgets.py
Czaki: It is not Idea. It is knowledge created in pain. 
You may also see this error in tests executed on main. 
Will try to fix this. 
In general, the problem comes from previous tests that are spawning the Viewer. 

in a general `make_napari_viewer` fixture has few weak spots (but writing a test without using it, may lead to even more spectacular failures) 

I even have opened some PR to fix it, but it is really time-consuming. 
for example https://github.com/napari/napari/pull/6745

src/napari_spatialdata/_scatterwidgets.py
```diff
@@ -374,3 +202,846 @@ def clear(self) -> None:
...
+
+        for i, (x, y) in enumerate(zip(self.scatter.xData, self.scatter.yData)):
+            point = Point(x, y)
+            # Check if the point belongs to any ROI
+            for polygon in polygon_list:
+
+                if polygon.contains(point):
+                    boolean_vector[i] = True
+                    break
+
```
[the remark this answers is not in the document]
fjorka: Fair point. Added it to a list of future improvements.

tests/test_scatterwidgets.py
fjorka: I'm not sure I understand your question. In this test the 'Plot' button is triggered while the widget holds discrete data. The test just checks that the control is created for discrete and not continuous color.

tests/test_scatterwidgets.py
fjorka: Pyqtgraph provides a default AutoRange functionality (small A icon in the left bottom corner) that just resets the axis.
The home button is providing a reset to the original way the graph is plotted (both AutoRange) and reset of the colors.
In retrospect calling it auto range was not the best choice but as the other one is default and not part of this implementation, there should not be a possibility of confusing them.

I added comments in the test.

src/napari_spatialdata/_view.py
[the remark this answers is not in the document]
fjorka: Very good point and thank you for a detailed explanation! Fixed in 019c5189de0378ba62ddc82bc449e9dadf2caf5d

tests/test_scatterwidgets.py
fjorka: The [0] in plot_widget.scatter_plot.scene().views()[0].viewport() refers to the first QGraphicsView in the list of views associated with the QGraphicsScene. As in case of a single graph there is only one view associated with the scene, this will not cause problems.

fjorka:
The notebooks start with the instruction:
```
There are two options to install napari-spatialdata:

(1) Run `pip install napari-spatialdata`
or,
(2) Clone this [repo](https://github.com/scverse/napari-spatialdata) and run `pip install -e .`
```
However, following this instruction and starting napari (or scatter widget) will give an error `ImportError: No Qt bindings could be found.` What about asking explicitly for napari installation first  d40fb7cec410b6f8a0778999d210d81b28eeffca which will inform a user about `Qt`?

src/napari_spatialdata/_view.py
```diff
@@ -116,12 +171,167 @@ def __init__(self, napari_viewer: Viewer, model: DataModel | None = None):
...
+            file_path = d.show_dialog()
+
+            if file_path:
+                if file_path.endswith(".h5ad") or file_path.endswith(".zarr") or file_path.endswith(".csv"):
+                    if file_path.endswith(".h5ad"):
+                        self.model.adata.write(file_path)
+                    if file_path.endswith(".zarr"):
+                        self.model.adata.write_zarr(file_path)
+                    if file_path.endswith(".csv"):
+                        self.model.adata.write_csvs(file_path)
```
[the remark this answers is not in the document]
fjorka: Let's bring also this comment to the conversation because initiating and saving are related:

> I would only make one thing clearer in the "scatterwidget.ipynb" and "scatterwidget_annotation.ipynb" notebooks.
I would clarify a bit around the scenarios: "napari is running" vs "no napari viewer detected" and the case in which the table is from a standalone AnnData vs coming from a SpatialData object.
In principle there are 4 cases. Are these two possible: "napari is running" and "the table is from a standalone AnnData" and "no napari viewer detected" and "the table is from a SpatialData object?

For simplicity, I suggest that we provide the user with only two options:

1. Running the Scatter Widget with AnnData.
2. Running Interactive mode with sdata.

For option 1, we do not specify the source of the AnnData, as we currently do not support modification of an existing AnnData file. This will hopefully be revisited in the future when such functionality becomes available. AnnData can either come directly from a file or be derived from an sdata structure, but in both cases, it is detached from its origin. It can be saved as a new file, or the user may overwrite the existing AnnData file (after acknowledging the OS's generic warning—AnnDataSaveDialog provides this functionality). If the AnnData is derived from sdata, it is the user's responsibility to reattach the modified table to the original sdata, if desired. If we keep this behavior, the `AnnDataSaveDialog` has a totally different purpose than `SaveDialog`.

If a user wishes to interact with the sdata object, it should be done through Interactive mode and not solely through the Scatter Widget. I don't see any downside to this approach, but I am curious about your opinion.

For option 2, when the Scatter Widget is used as part of the Interactive viewer, it obtains sdata from the layer on which it is called. The status of the Scatter Widget indicates whether the sdata from the layer is backed. When a user clicks 'Save', the sdata will be overwritten if backed; otherwise, a logger will display the message: "Spatialdata object is not stored on disk, can only add element in memory."

On a technical level:

The Annotation button action does not depend on the sdata status. When the button is clicked:
- The AnnData of the Scatter Widget is modified.
- The AnnData of the layer is modified.
- The sdata of the layer is modified (a merge of the table is performed).

The 'Save' button calls `_write_element_to_disk` (similar to the AnnotationWidget), which is a convenient place to add additional precautions, such as saving a temporary table and renaming it later.

Since the Scatter Widget always works with an existing table, overwriting a table in sdata is unavoidable. Given that every click of the 'Save' button would result in an overwrite, should we consider adding a warning to the notebooks and documentation instead of displaying a dialog box? This might be a better approach, as there isn’t much the user can do to make the process safer at this point. What do you think?

fjorka:
To do:

- [ ] new screenshots for the notebooks (after approving the new look by @LucaMarconato )
- [ ] add explanations in `scatterwidget.ipynb` and `scatterwidget_annotation.ipynb` about starting with AnnData and options for saving
- [ ] (possibly) add a warning dialog about overwriting sdata object

LucaMarconato:
TODO for me: dataset

src/napari_spatialdata/_view.py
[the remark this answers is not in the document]
LucaMarconato: (the conversation can be resolved)

LucaMarconato:
Thanks @fjorka and @melonora, we finished reviewing and adjusting some last open points, and now we are good to merge!

Really excited to ultimately release this! We will include this PR in a release X.com post that we will make in the near future.

I will consolidate the points open for follow up PRs in a separate issue.

LucaMarconato (approved):

LucaMarconato:
@fjorka @melonora I forgot about this file https://github.com/fjorka/napari-spatialdata/blob/7d60e030aff750ee08fd874b8c6b6425ec5fff79/tests/test_aaa.py, which if I got it right, was a work-in-progress implementation for an alternative approach to span the test napari viewer, in order to avoid the segfault problem with Python 3.9. I think the problem have disappeared due to having dropped the Python 3.9 support. So please @melonora in the follow up PR can you drop that file? Thanks.

LucaMarconato:
Actually there is still a problem with tests, I reported it here https://github.com/scverse/napari-spatialdata/issues/335 so we can continue the discussion there. CC @melonora

PR #276, opened 2024-07-19
fix: fix adjusting contrast limits for rgb data

Czaki:
While working on reproducing of performance problem for points and shapes, I found that adjusting contrast limits do not work for data with 3 channels, that are represented as RGB data. 

This PR contains the following changes:

1) change `channel_changed` from local function to class method 
2) generalize `_channel_changed` to accept any dimensional data. 
3) fix `AnnData` initialization for rgb data

### DIFF
--- a/src/napari_spatialdata/_view.py
+++ b/src/napari_spatialdata/_view.py
@@ -300,7 +300,7 @@ class QtAdataViewWidget(QWidget):
         has_adata = layer is not None and layer.metadata.get("adata") is not None
 
         # has_adata is added so we see the channels in the view widget under vars
-        if layer is None or not is_image or not has_sdata or not has_adata:
+        if layer is None or not is_image or not has_sdata or not has_adata or layer.rgb:
             return
 
         current_point = list(event.value)
--- a/src/napari_spatialdata/_view.py
+++ b/src/napari_spatialdata/_view.py
@@ -238,37 +238,7 @@ class QtAdataViewWidget(QWidget):
         self.var_widget = AListWidget(self.viewer, self.model, attr="var")
         self.var_widget.setAdataLayer("X")
 
-        def channel_changed(event: Event) -> None:
-            layer = self.model.layer
-            is_image = isinstance(layer, Image)
-
-            has_sdata = layer is not None and layer.metadata.get("sdata") is not None
-            has_adata = layer is not None and layer.metadata.get("adata") is not None
-
-            # has_adata is added so we see the channels in the view widget under vars
-            if layer is not None and is_image and has_sdata and has_adata:
-                c_channel = event.value[0]
-
-                # TODO remove once contrast limits in napari are fixed
-                if isinstance(layer.data, MultiScaleData):
-                    # just compute lowest resolution
-                    image = layer.data[-1][c_channel, :, :].compute()
-                    min_value = image.min().data
-                    max_value = image.max().data
-                else:
-                    image = layer.data[c_channel, :, :].compute()
-                    min_value = image.min()
-                    max_value = image.max()
-                if min_value == max_value:
-                    min_value = np.iinfo(image.data.dtype).min
-                    max_value = np.iinfo(image.data.dtype).max
-                layer.contrast_limits = [min_value, max_value]
-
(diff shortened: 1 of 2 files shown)


src/napari_spatialdata/_view.py
```diff
@@ -322,6 +292,49 @@ def channel_changed(event: Event) -> None:
...
 
+    def _channel_changed(self, event: Event) -> None:
+        layer = self.model.layer
+        is_image = isinstance(layer, Image)
+
+        has_sdata = layer is not None and layer.metadata.get("sdata") is not None
+        has_adata = layer is not None and layer.metadata.get("adata") is not None
+
+        # has_adata is added so we see the channels in the view widget under vars
+        if layer is None or not is_image or not has_sdata or not has_adata:
```
Czaki, line 303: maybe it should look like 
```suggestion
        if layer is None or not is_image or not has_sdata or not has_adata or image.rgb:
```

src/napari_spatialdata/_view.py
```diff
@@ -322,6 +292,49 @@ def channel_changed(event: Event) -> None:
...
+            return
+
+        current_point = list(event.value)
+        displayed = self._viewer.dims.displayed
+
+        for i, (lo_size, hi_size, cord) in enumerate(zip(layer.data[-1].shape, layer.data[0].shape, displayed)):
+            if i in displayed:
+                current_point[i] = slice(None)
+            else:
+                current_point[i] = int(cord * lo_size / hi_size)
```
Czaki, line 313: This is logic for any dimensional data. 

If you are 100% sure that data will always be 2D, then this logic may be obsolete. 
If contrast limits needs to be calculated for the whole channel, not only the visible portion, then it is also overcomplicated. But then, for 3D data, should be no recalculation when change Z slice.

melonora: In the future we will not be sure about this. SpatialData already does have 3D models. Where bugs occur for 3D napari-spatialdata should be adjusted.

melonora:
I will first add #153 . That PR required some upstream changes that I have now implemented. I will then check how your PR interacts with it.

src/napari_spatialdata/_viewer.py
```diff
@@ -434,7 +434,8 @@ def add_sdata_image(self, sdata: SpatialData, key: str, selected_cs: str, multi:
         affine = _get_transform(sdata.images[original_name], selected_cs)
         rgb_image, rgb = _adjust_channels_order(element=sdata.images[original_name])
 
-        channels = get_channels(sdata.images[original_name])
+        channels = (0,) if rgb else get_channels(sdata.images[original_name])
```
melonora, line 437:
```suggestion
        channels = ("RGB(A),) if rgb else get_channels(sdata.images[original_name])
```
With #153 we have different behaviour now for detecting whether something is supposed to be RGB(A). When parsing Image2D/3DModel users can specify c_coords as ('r','g','b') with optionally alpha to indicate it is a RGB image. I would prefer to just have the channel specified as `RGB(A)` in that case similar to colormap of the napari image layers when it is specified to be a RGB image.

src/napari_spatialdata/_view.py
melonora, line 303:
```suggestion
        if layer is None or not is_image or not has_sdata or not has_adata or layer.rgb:
```

melonora (changes requested):
Just minor changes to take into account the changes made in #153. Thanks @Czaki !

src/napari_spatialdata/_viewer.py
```diff
@@ -434,7 +434,8 @@ def add_sdata_image(self, sdata: SpatialData, key: str, selected_cs: str, multi:
         affine = _get_transform(sdata.images[original_name], selected_cs)
         rgb_image, rgb = _adjust_channels_order(element=sdata.images[original_name])
 
-        channels = get_channels(sdata.images[original_name])
+        channels = ("RGB(A),) if rgb else get_channels(sdata.images[original_name])
```
Czaki, line 437:
```suggestion
        channels = ("RGB(A)",) if rgb else get_channels(sdata.images[original_name])
```

PR #279, opened 2024-07-29
docs: Fix installation instruction

Czaki:
Since a long time, the PyQt5 wheels have been available on PyPi. This part of instruction is no longer relevant. 

Also specify PyQt5 dependency using `napari[pyqt5]` to enforce version constraints from napari (to not install incompatible version of `PyQt5`)


LucaMarconato:
Thank you, much cleaner now!

PR #285, opened 2024-08-08
Sort elements in widgets

### LINKED ISSUE #284: Coordinate systems widget not sorted
Coordinate systems are displayed non-alphabetically. With a large number of coordinate systems, this makes it hard to find and select a specific one.
Screenshot

If things are unordered, it usually can be caused by the order of file loading (glob on Unix is unordered / inode order, but alphabetical on Windows).

Although elements display alphabetically, there is nothing in the source code that sorts them (they are ordered just by chance, probably order of insertion to the dictionary).

## Suggestion

- Sort elements before adding them to the elements widget
- Sort coordinate systems before adding them to the widget
- Always put "global" first (out of alphabetical order), since it is the default coordinate system and users may want to access it easily. Strict alphabetical order would put it somewhere in the middle, harder to find.

aeisenbarth:
Closes #284

### DIFF
--- a/src/napari_spatialdata/_sdata_widgets.py
+++ b/src/napari_spatialdata/_sdata_widgets.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from operator import itemgetter
 from pathlib import Path
 from typing import TYPE_CHECKING, Iterable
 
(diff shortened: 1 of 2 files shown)


LucaMarconato:
LGTM thanks! I would merge after https://github.com/scverse/napari-spatialdata/pull/270 since that PR makes substantial changes to the scatter widget and the test that fails is related to the scatter plot widget.

LucaMarconato:
Thanks @aeisenbarth for the fix!

PR #291, opened 2024-08-15
Modernize release workflow use most recent tool

Czaki:
This PR is modernizing release workflow (and allowing for future switch into trusted publishers) to use most recent tools best practices. 

The main missing thing is to cut information from the changelog file and provide it as a body.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,30 +1,39 @@
 name: Release
 
 on:
-  release:
-    types: [published]
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - "v*" # Push events to matching v*, i.e. v1.0, v20.15.10
+  pull_request:
+    paths:
+      - .github/workflows/make_release.yml
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,30 +1,39 @@
...
+        run: echo "tag=${GITHUB_REF/refs\/tags\/v/}" >> "$GITHUB_ENV"
+      - name: Create Release
+        uses: "softprops/action-gh-release@v2"
+        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
+        with:
+          tag_name: ${{ github.ref }}
+          name: ${{ env.tag }}
+          draft: false
+          prerelease: ${{ contains(env.tag, 'rc') || contains(env.tag, 'a') || contains(env.tag, 'b') }}
+          target_commitish: ${{ github.sha }}
```
Czaki, line 32: @LucaMarconato This line will save proper information in release to be later used by benchmarks.

PR #299, opened 2024-08-19
Minimum 3.9.2, testing against 3.11

LucaMarconato:
While debugging some code against `python` 3.9, I noticed how using `python` 3.9.0 or `3.9.1` systematically leads to these bugs https://github.com/dask/distributed/issues/7956.

Generally one would install a later patch version, but in this PR I add the constraint explicitly in the dependencies. 

Also I did some cleanup:
- black had `target-version = 'py38'`, I changed it to `py39`
- I am testing against 3.11 instead of 3.10. We could also test against 3.10 instead of only the min-max versions, at the cost of more CI time.


LucaMarconato:
General housekeeping, merging.

PR #301, opened 2024-08-19
fix #282 (error in channel selection due to #276)

### LINKED ISSUE #282: Selecting individual channels throws a ValueError
Hi,

I tried to select individual channels, but keep getting a bug along the lines of `ValueError: applied function returned data with an unexpected number of dimensions.`. For 3-channel images, the bug does not occur, but it is also not possible to select channels on the right side in this case.

Minimal example:
```
import numpy as np
import spatialdata

# works, but can't select any channels
image = spatialdata.models.Image2DModel.parse(
    np.random.randint(low=0, high=255, size=(3, 3000, 3000)), dims=("c", "x", "y")
)
sd_obj = spatialdata.SpatialData(images={"image": image})

# does not work
image = spatialdata.models.Image2DModel.parse(
    np.random.randint(low=0, high=255, size=(5, 3000, 3000)), dims=("c", "x", "y")
)
sd_obj = spatialdata.SpatialData(images={"image": image})

interactive = Interactive(sd_obj)
interactive.run()
```

LucaMarconato:
Closes #282 

The bug is now covered by a test.

### DIFF
--- a/src/napari_spatialdata/_view.py
+++ b/src/napari_spatialdata/_view.py
@@ -6,6 +6,7 @@ from typing import Any
 import napari
 import numpy as np
 import pandas as pd
+import xarray
 from anndata import AnnData
 from loguru import logger
 from matplotlib.colors import to_rgba_array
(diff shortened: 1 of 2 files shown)


LucaMarconato:
The failing test is due to this https://github.com/napari/napari/issues/7208.

PR #303, opened 2024-08-21
test to reproduce #302

### LINKED ISSUE #302: Error with channel slider due to #276
I have found another problem that appeared after #276 (this other problem, https://github.com/scverse/napari-spatialdata/issues/282 is now fixed).

I'll open a PR with tests to reproduce the problem.

LucaMarconato:
closes #302 

(no fix in this PR yet, for the moment I just reproduce the bug)

### DIFF
--- a/src/napari_spatialdata/_view.py
+++ b/src/napari_spatialdata/_view.py
@@ -307,12 +307,16 @@ class QtAdataViewWidget(QWidget):
 
         current_point = list(event.value)
         displayed = self._viewer.dims.displayed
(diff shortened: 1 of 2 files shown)


LucaMarconato:
This is the failing test (as expected):
`tests/test_view.py::test_channel_slider_images_bug_302[QtAdataViewWidget-False] FAILED [ 66%]`.

@Czaki would you have time to look at this? It fails on a part of code that you wrote in #302, so probably it would be easier for you to address it.

Czaki:
I do not notice single sclae case. It should be ready.

PR #304, opened 2024-08-23
bug categorical annotatoins points

LucaMarconato:
Fix PR:
- fixes a bug when trying to plot categorical annotations, stored in the dataframe
- adds tests for points, both for categorical and continuous annotations stored in the dataframe column
- allows for categorical colors to be stored in the dataframe column <- **new feature**


LucaMarconato:
Ok to merge since the failing test is on the matplotlib widget, which will be replaced by the new code on the scatterplot widget.

PR #317, opened 2024-09-25
Fix bug adata in model not being reset

LucaMarconato:
Fixes two bugs:
- there was a hardcoded `genes_color`, I fixed this
- if an element with a table was added and removed, the `mode.table` was not being reset, leading to inconsistent behavior

CC @melonora


melonora (approved):
LGTM!

LucaMarconato:
No `pre-commit` problems locally, let's fix later.

PR #318, opened 2024-09-27
fix radii of transformed circles

LucaMarconato:
The calculation for the radii of transformed circles was wrong. This was manifested for instance when switching between the `global` and `downscaled_hires` coordinate systems in Visium.


LucaMarconato:
Small PR. But I also fixed the pre-commits (most of the code changes).

PR #345, opened 2025-02-06
Fix exception due to adata being none

LucaMarconato:
CI is red because of this: https://github.com/scverse/napari-spatialdata/pull/343.

The PR fixes two exception that originated when an annotating table `adata` was `None`. I discovered this bug with this code snippet: https://github.com/scverse/spatialdata/pull/864.


LucaMarconato:
The tests that fail are:
- benchmark (due to `libmamba`), will fix separately
- failed test due to object not cleaned-up. Flaring up randomly, we need to address it separately.

PR #347, opened 2025-02-13
Temporarily constrain napari version

melonora:
With napari version 0.5.6 there have been collisions in the shortcuts being used for both inheriting layer metadata and exporting a layer to an sdata object. This PR temporarily constrains the napari version being used.


melonora:
After discussing with @LucaMarconato , we decided to merge already as the benchmark test is not related to this PR and the change is minimal and not thought to effect the benchmark in the first place.

PR #357, opened 2025-04-21
fix tests cleanup

LucaMarconato:
Even after the [latest fix](https://github.com/scverse/napari-spatialdata/pull/349), some [tests sometimes fail](https://github.com/scverse/napari-spatialdata/actions/runs/14579292539/job/40892220813) due to improper cleanup. Probably because the fix was not extended to those tests. 

In this PR, I fix only the test that was failing in the CI run above. Two questions for you:
- should we extend this to all the tests in `test_spatialdata.py`?
- IIRC `test_aaa.py` was an experiment to try to fix the sporadically broken tests, and was following some suggestions from [here](https://github.com/scverse/napari-spatialdata/pull/270#discussion_r1731638385). If your fix works, I think we can remove that test. WDYT?

### DIFF
--- a/tests/test_spatialdata.py
+++ b/tests/test_spatialdata.py
@@ -11,6 +11,7 @@ from dask.dataframe import from_dask_array
 from multiscale_spatial_image import to_multiscale
 from napari.layers import Image, Labels, Points
 from napari.utils.events import EventedList
+from napari.viewer import Viewer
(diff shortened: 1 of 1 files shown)


LucaMarconato:
I'll remove `test_aaa.py`. I will not extend the change of this PR to all the test unless needed.

PR #364, opened 2025-06-02
Add channel widget

melonora:
Users currently don't have the option to select to load individual channels. While the table widget is nice for when working with rasterizations that don't result in large images, for spatial proteomics data in which each channel is large, this is not a nice experience.
This PR presents a first way of improving this.

The channel element widget has the following characteristics:

- It updates upon selection chance in the element widget
- It has no issues with multiple spatialdata objects as it is based on current selected element
- When adding a layer this way, there is a suffix `_ch:channel_name`.

In addition to that, this PR has removed the colorbar slider in order to make more space for the new widget.

Lastly, this PR fixes a bug of coordinate systems in the `CoordinateSystemWidget` not being de-duplicated in case of loading multiple `SpatialData` objects.


melonora:
Had to pin dask as the legacy is not supported anymore in the newer versions

PR #368, opened 2025-07-03
Update napari to 0.6.2 and add bermuda

melonora:
With this PR the plugin uses only napari 0.6.2 and up. Also, bermuda can optionally be installed. Next, the command for linking layers has been changed to 'Ctrl-L' (command key on MacOS). This, because 'Shift-L' is now used to switch to poly line drawing in the 'Shapes' layer.
Lastly, the deprecated access of "_dock_widgets" has been changed to accessing the public property in order to silence the warning.


src/napari_spatialdata/_view.py
```diff
@@ -426,7 +426,7 @@ def __init__(self, napari_viewer: Viewer, model: DataModel | None = None) -> Non
         super().__init__()
 
         self._viewer = napari_viewer
-        self._model = model if model else napari_viewer.window._dock_widgets["SpatialData"].widget().viewer_model._model
+        self._model = model if model else napari_viewer.window.dock_widgets["SpatialData"].viewer_model.model
```
TimMonko: use a `hasattr` to make compatible with older napari versions, if you want

PR #366, opened 2025-06-22
Better 3D and 2.5D support for raster and vector data

LucaMarconato:
Work in progress, developed together with https://github.com/EricMoerthVis/tissue-map-tools/pull/10

PR #393, opened 2026-04-15
Support 3D visualization for points and 2.5D shapes

### LINKED ISSUE #31: 3D visualization
Support 3D visualization for points, temporarily disabled.

asarigun:
Closes #31.

Builds on #366 to complete 3D support for vector data:

- 2.5D circles rendered at correct z depth
- Affine transforms respect z axis on coordinate system switch
- Save round-trip no longer drops z from points/shapes
- UI toggles for 3D points and 2.5D shapes projection
- Z-range slider to filter by depth

### DIFF
--- a/src/napari_spatialdata/_sdata_widgets.py
+++ b/src/napari_spatialdata/_sdata_widgets.py
@@ -15,11 +15,22 @@ from napari.utils.notifications import show_info
 from packaging.version import parse as parse_version
 from qtpy.QtCore import QThread, Signal
 from qtpy.QtGui import QIcon
-from qtpy.QtWidgets import QLabel, QListWidget, QListWidgetItem, QProgressBar, QVBoxLayout, QWidget
+from qtpy.QtWidgets import (
+    QCheckBox,
+    QHBoxLayout,
+    QLabel,
+    QListWidget,
+    QListWidgetItem,
+    QProgressBar,
+    QVBoxLayout,
+    QWidget,
+)
+from superqt import QDoubleRangeSlider
 from spatialdata import SpatialData
 from spatialdata.models._utils import DEFAULT_COORDINATE_SYSTEM
 
 from napari_spatialdata._viewer import SpatialDataViewer
+from napari_spatialdata.constants import config
 from napari_spatialdata.constants.config import N_CIRCLES_WARNING_THRESHOLD, N_SHAPES_WARNING_THRESHOLD
 from napari_spatialdata.utils._utils import _get_sdata_key, get_duplicate_element_names, get_elements_meta_mapping
 
@@ -174,11 +185,45 @@ class SdataWidget(QWidget):
         self.slider.setRange(0, 0)
         self.slider.setVisible(False)
 
+        self.enable_3d_points = QCheckBox("Enable 3D points")
+        self.enable_3d_points.setChecked(not config.PROJECT_3D_POINTS_TO_2D)
+        self.enable_3d_points.setToolTip("When checked, points with a z coordinate are displayed in 3D.")
+        self.enable_3d_points.toggled.connect(self._on_3d_points_toggled)
+
+        self.enable_2_5d_shapes = QCheckBox("Enable 2.5D shapes")
+        self.enable_2_5d_shapes.setChecked(not config.PROJECT_2_5D_SHAPES_TO_2D)
+        self.enable_2_5d_shapes.setToolTip("When checked, shapes with a z coordinate are displayed in 2.5D.")
+        self.enable_2_5d_shapes.toggled.connect(self._on_2_5d_shapes_toggled)
+
+        self.z_range_label = QLabel("Z range:")
+        self.z_range_value_label = QLabel("")
+        z_range_header = QHBoxLayout()
+        z_range_header.addWidget(self.z_range_label)
+        z_range_header.addWidget(self.z_range_value_label)
+        z_range_header.addStretch()
+        self.z_range_header_widget = QWidget()
+        self.z_range_header_widget.setLayout(z_range_header)
+
+        self.z_range_slider = QDoubleRangeSlider()
+        self.z_range_slider.setRange(0.0, 1.0)
+        self.z_range_slider.setValue((0.0, 1.0))
+        self.z_range_slider.setToolTip("Filter visible points and shapes by z coordinate range.")
+        self.z_range_slider.valueChanged.connect(self._on_z_range_changed)
+
+        self._z_slider_visible = False
+        self.z_range_header_widget.setVisible(False)
+        self.z_range_slider.setVisible(False)
+
         self.layout().addWidget(self.slider)
         self.layout().addWidget(QLabel("Coordinate System:"))
         self.layout().addWidget(self.coordinate_system_widget)
         self.layout().addWidget(QLabel("Elements:"))
         self.layout().addWidget(self.elements_widget)
+        self.layout().addWidget(QLabel("3D Settings:"))
+        self.layout().addWidget(self.enable_3d_points)
+        self.layout().addWidget(self.enable_2_5d_shapes)
+        self.layout().addWidget(self.z_range_header_widget)
+        self.layout().addWidget(self.z_range_slider)
         self.elements_widget.itemDoubleClicked.connect(self._on_click_item)
         self.coordinate_system_widget.currentItemChanged.connect(
             lambda item: self.elements_widget._onItemChange(item.text())
@@ -196,12 +241,14 @@ class SdataWidget(QWidget):
     def _on_insert_layer(self, event: Event) -> None:
         layer = event.value
         layer.events.visible.connect(self._update_visible_in_coordinate_system)
+        self._update_z_slider()
 
     def _on_click_item(self, item: QListWidgetItem) -> None:
         self._onClick(item.text())
 
     def _hide_slider(self) -> None:
         self.slider.setVisible(False)
+        self._update_z_slider()
 
     def _onClick(self, text: str) -> None:
         selected_cs = self.coordinate_system_widget._system
@@ -257,6 +304,38 @@ class SdataWidget(QWidget):
                         # Prevent _update_visible_in_coordinate_system of invalid removal of coordinate system
                         layer.metadata["_active_in_cs"].add(coordinate_system)
                         layer.metadata["_current_cs"] = coordinate_system
+
+    def _on_3d_points_toggled(self, checked: bool) -> None:
+        config.PROJECT_3D_POINTS_TO_2D = not checked
+
+    def _on_2_5d_shapes_toggled(self, checked: bool) -> None:
+        config.PROJECT_2_5D_SHAPES_TO_2D = not checked
+
+    def _update_z_slider(self) -> None:
+        """Show the z-range slider when layers with z data are present and update its range."""
+        z_range = self.viewer_model.get_z_range()
+        if z_range is None:
+            self.z_range_header_widget.setVisible(False)
+            self.z_range_slider.setVisible(False)
+            self._z_slider_visible = False
+            return
+
+        z_min, z_max = z_range
+        if z_min == z_max:
+            z_max = z_min + 1.0
+
+        self.z_range_slider.setRange(z_min, z_max)
+        if not self._z_slider_visible:
+            self.z_range_slider.setValue((z_min, z_max))
+            self._z_slider_visible = True
+        self.z_range_value_label.setText(f"[{z_min:.1f}, {z_max:.1f}]")
+        self.z_range_header_widget.setVisible(True)
+        self.z_range_slider.setVisible(True)
+
+    def _on_z_range_changed(self, value: tuple[float, float]) -> None:
+        z_min, z_max = value
+        self.z_range_value_label.setText(f"[{z_min:.1f}, {z_max:.1f}]")
+        self.viewer_model.filter_layers_by_z_range(z_min, z_max)
 
     def _get_shapes(self, sdata: SpatialData, key: str, selected_cs: str, multi: bool) -> Shapes | Points:
         original_name = key[: key.rfind("_")] if multi else key
--- a/src/napari_spatialdata/_sdata_widgets.py
+++ b/src/napari_spatialdata/_sdata_widgets.py
@@ -25,9 +25,9 @@ from qtpy.QtWidgets import (
     QVBoxLayout,
     QWidget,
 )
-from superqt import QDoubleRangeSlider
 from spatialdata import SpatialData
 from spatialdata.models._utils import DEFAULT_COORDINATE_SYSTEM
+from superqt import QDoubleRangeSlider
 
 from napari_spatialdata._viewer import SpatialDataViewer
 from napari_spatialdata.constants import config
--- a/src/napari_spatialdata/_sdata_widgets.py
+++ b/src/napari_spatialdata/_sdata_widgets.py
@@ -17,7 +17,6 @@ from qtpy.QtCore import QThread, Signal
 from qtpy.QtGui import QIcon
 from qtpy.QtWidgets import (
     QCheckBox,
-    QHBoxLayout,
     QLabel,
     QListWidget,
     QListWidgetItem,
@@ -27,7 +26,6 @@ from qtpy.QtWidgets import (
 )
 from spatialdata import SpatialData
 from spatialdata.models._utils import DEFAULT_COORDINATE_SYSTEM
-from superqt import QDoubleRangeSlider
 
 from napari_spatialdata._viewer import SpatialDataViewer
 from napari_spatialdata.constants import config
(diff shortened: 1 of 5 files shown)


src/napari_spatialdata/utils/_utils.py
```diff
@@ -462,10 +462,13 @@ def generate_random_color_hex() -> str:
 def _get_ellipses_from_circles(yx: ArrayLike, radii: ArrayLike) -> ArrayLike:
     """Convert circles to ellipses.
 
+    Supports both 2D (y, x) and 2.5D (z, y, x) centroids. For 2.5D input the radius is
```
LucaMarconato: thanks for adding the docstring. I'd take the opportunity to rename yx to coords.

src/napari_spatialdata/utils/_utils.py
```diff
@@ -462,10 +462,13 @@ def generate_random_color_hex() -> str:
...
+    applied only to y and x while z is kept constant across the four corner vertices.
+
     Parameters
     ----------
     yx
-        Centroids of the circles.
+        Centroids of the circles with shape ``(N, 2)`` or ``(N, 3)``.
```
LucaMarconato: Good to mention again the order yx or zyx.

src/napari_spatialdata/utils/_utils.py
```diff
@@ -475,13 +478,27 @@ def _get_ellipses_from_circles(yx: ArrayLike, radii: ArrayLike) -> ArrayLike:
         Ellipses.
     """
     ndim = yx.shape[1]
-    assert ndim == 2
-    r = np.stack([radii] * ndim, axis=1)
-    lower_left = yx - r
-    upper_right = yx + r
+    assert ndim in (2, 3)
```
LucaMarconato: please replace with if -> raise, to avoid `python -O` to bypass the assert.

src/napari_spatialdata/_sdata_widgets.py
```diff
@@ -174,11 +185,45 @@ def __init__(self, viewer: Viewer, sdata: EventedList):
...
 
+        self.enable_3d_points = QCheckBox("Enable 3D points")
+        self.enable_3d_points.setChecked(not config.PROJECT_3D_POINTS_TO_2D)
+        self.enable_3d_points.setToolTip("When checked, points with a z coordinate are displayed in 3D.")
+        self.enable_3d_points.toggled.connect(self._on_3d_points_toggled)
+
+        self.enable_2_5d_shapes = QCheckBox("Enable 2.5D shapes")
+        self.enable_2_5d_shapes.setChecked(not config.PROJECT_2_5D_SHAPES_TO_2D)
+        self.enable_2_5d_shapes.setToolTip("When checked, shapes with a z coordinate are displayed in 2.5D.")
+        self.enable_2_5d_shapes.toggled.connect(self._on_2_5d_shapes_toggled)
```
LucaMarconato: This is a great idea! Much easier for people working with 3D data to switch between 2D and 3D without having to go through the settings. 

A few comments:
1. many people still work with 2D data and these settings would take some screen real estate. It may be worth considering to show them only if at least one of the elements contains the z dimension.
2. it would be more correct to call the checkboxes "discard z for 3D points" and "discard z for 2.5D shapes", since the 2D vs 3D view is handled by napari, and the difference is if we drop or not z. If we go with the new naming, or an equivalent one, the checkboxes would be enabled by default.
3. it would be nice to make it clear to the users that the checkboxes apply to new layers, not on data that is already displayed. Not sure what is the best way to communicate this. "discard z for new 3D points" could be a way, but it's super long, and the original "Enable 3D points", even if not entirely correct, sounds much nicer and more intuitive.

LucaMarconato: Re point 3: definetely the tooltips should include a "new", to explain that only newly created layers respect the setting.

src/napari_spatialdata/_sdata_widgets.py
```diff
@@ -174,11 +185,45 @@ def __init__(self, viewer: Viewer, sdata: EventedList):
...
+        z_range_header.addWidget(self.z_range_value_label)
+        z_range_header.addStretch()
+        self.z_range_header_widget = QWidget()
+        self.z_range_header_widget.setLayout(z_range_header)
+
+        self.z_range_slider = QDoubleRangeSlider()
+        self.z_range_slider.setRange(0.0, 1.0)
+        self.z_range_slider.setValue((0.0, 1.0))
+        self.z_range_slider.setToolTip("Filter visible points and shapes by z coordinate range.")
+        self.z_range_slider.valueChanged.connect(self._on_z_range_changed)
```
LucaMarconato: The z range doesn't seem to work on the test data I am using (some Xenium data and some artificial data). I would nevertheless be inclined not to include it in this PR and consider adding this later. In this way the PR stays focused on extending the 3D viz to unsupported elements (such as circles) and making the settings more accessible, without introducing new data visualization strategies. 

In particular, before adding this slider we would definitely need to fix the broken value slide (the one with the viridis cmap), which has been non-functional already for a while.

src/napari_spatialdata/_sdata_widgets.py
```diff
@@ -256,6 +303,38 @@ def _update_layers_visibility(self) -> None:
...
+        z_range = self.viewer_model.get_z_range()
+        if z_range is None:
+            self.z_range_header_widget.setVisible(False)
+            self.z_range_slider.setVisible(False)
+            self._z_slider_visible = False
+            return
+
+        z_min, z_max = z_range
+        if z_min == z_max:
+            z_max = z_min + 1.0
```
LucaMarconato: I think that the +1 makes the min and max very close, in particular for 3D data thas has a large z-span. 

An increment as a fraction of the z-range may be better. Anyway, as explained in the previous comment, I would consider removing the code from this PR and then eventually moving it in a new PR.

src/napari_spatialdata/_sdata_widgets.py
LucaMarconato: Maybe the z-slicer is something that could live in a separate plugin, or even directly in napari. This would make it more accessible and keep napari-spatialdata lean and focused.

src/napari_spatialdata/_viewer.py
```diff
@@ -804,8 +812,113 @@ def _affine_transform_layers(self, coordinate_system: str) -> None:
                 sdata = metadata["sdata"]
                 element_name = metadata["name"]
                 element_data = sdata[element_name]
-                affine = _get_transform(element_data, coordinate_system)
+                include_z = self._should_include_z(element_data)
+                affine = _get_transform(element_data, coordinate_system, include_z=include_z)
```
LucaMarconato, line 816: I realized now that it's better to clarify that in `_get_transform()` the z is included not only if include_z is True, but also if the z is present in the transformation. This could be added in the docstring of _get_transform()

src/napari_spatialdata/_viewer.py
```diff
@@ -804,8 +812,113 @@ def _affine_transform_layers(self, coordinate_system: str) -> None:
...
                 if affine is not None:
                     layer.affine = affine
                     if layer._type_string == "points":
                         self._adjust_radii_of_points_layer(layer, affine)
+
+    @staticmethod
+    def _should_include_z(element: DaskDataFrame | GeoDataFrame) -> bool:
```
LucaMarconato, line 838: Looks good! I would call this functionn for each element in the `sdatas` objects being passed to interactive to determine if showing the 2 checkboxes about considering the 3D data.

src/napari_spatialdata/_viewer.py
```diff
@@ -804,8 +812,113 @@ def _affine_transform_layers(self, coordinate_system: str) -> None:
...
+            if not metadata.get("sdata"):
+                continue
+            sdata = metadata["sdata"]
+            element_name = metadata["name"]
+            element_data = sdata[element_name]
+            axes = get_axes_names(element_data)
+            if "z" not in axes:
+                continue
+            if isinstance(element_data, DaskDataFrame):
+                z_vals = element_data["z"].compute().values
```
LucaMarconato: Expensive call. This is another reason to move the slider functionality outside this PR.

src/napari_spatialdata/_viewer.py
```diff
@@ -804,8 +812,113 @@ def _affine_transform_layers(self, coordinate_system: str) -> None:
...
+            if len(z_vals) == 0:
+                continue
+            found = True
+            z_min = min(z_min, float(z_vals.min()))
+            z_max = max(z_max, float(z_vals.max()))
+        if not found:
+            return None
+        return z_min, z_max
+
+    def filter_layers_by_z_range(self, z_min: float, z_max: float) -> None:
```
LucaMarconato: Neuroglancer let the user specify a shader to render points, which for instance include the functionality of this function. This makes custom rendering super fast. We could check, in a follow up work, if something similar is possible (by reaching out to the napari team).

tests/test_3d_visualization.py
```diff
@@ -111,6 +114,464 @@ def test_2_5d_shapes_full_3d(self, make_napari_viewer: Any, sdata_2_5d_shapes: S
...
+            viewer.layers.clear()
+
+            widget.enable_3d_points.setChecked(True)
+            widget._onClick("points_3d")
+            assert viewer.layers[0].data.shape[1] == 3
+        finally:
+            config.PROJECT_3D_POINTS_TO_2D = original_value
+
+
+class TestZBinning:
```
LucaMarconato: test for a follow up PR

tests/test_3d_visualization.py
```diff
@@ -111,6 +114,464 @@ def test_2_5d_shapes_full_3d(self, make_napari_viewer: Any, sdata_2_5d_shapes: S
             config.PROJECT_2_5D_SHAPES_TO_2D = original_value
 
 
+class Test2_5DCirclesVisualization:
```
LucaMarconato, line 90: looks good! One suggested change:
I realized that each of these paris of tests:
1. test_3d_points_projected_to_2d, test_3d_points_full_3d
2. test_2_5d_shapes_projected_to_2d, test_2_5d_shapes_full_3d
3. test_2_5d_circles_projected_to_2d, test_2_5d_circles_full_3d
can be moved into single tests (by using paramaterize). This would make the code more readable.

tests/test_3d_visualization.py
```diff
@@ -111,6 +114,464 @@ def test_2_5d_shapes_full_3d(self, make_napari_viewer: Any, sdata_2_5d_shapes: S
...
+            widget._onClick("circles_2.5d")
+
+            assert len(viewer.layers) == 1
+            # Full 3D: coordinates should have 3 values (z, y, x)
+            assert viewer.layers[0].data.shape[1] == 3
+        finally:
+            config.PROJECT_2_5D_SHAPES_TO_2D = original_value
+
+
+class TestAffineTransformLayers:
```
LucaMarconato, line 123: also the two tests in this class could be condensed into a single test by using parametrize (and changing a bit the code in the second test).

tests/test_3d_visualization.py
```diff
@@ -111,6 +114,464 @@ def test_2_5d_shapes_full_3d(self, make_napari_viewer: Any, sdata_2_5d_shapes: S
...
+            widget.coordinate_system_widget._select_coord_sys("scaled")
+            widget.viewer_model._affine_transform_layers("scaled")
+
+            # Projected to 2D -> affine stays 3x3
+            assert layer.affine.affine_matrix.shape == (3, 3)
+        finally:
+            config.PROJECT_3D_POINTS_TO_2D = original_value
+
+
+class TestSavePointsPreservesZ:
```
LucaMarconato, line 170: same also for the tests in TestSavePointsPreservesZ and TestSaveShapesPreservesZ, the 2 tests in each of the class can be condensed into one.

tests/test_3d_visualization.py
```diff
@@ -111,6 +114,464 @@ def test_2_5d_shapes_full_3d(self, make_napari_viewer: Any, sdata_2_5d_shapes: S
...
+
+            parsed, _ = widget.viewer_model._save_shapes_to_sdata(layer, "shapes_2.5d", overwrite=True)
+
+            saved_axes = get_axes_names(parsed)
+            assert "z" not in saved_axes
+        finally:
+            config.PROJECT_2_5D_SHAPES_TO_2D = original_value
+
+
+class TestUIToggle:
```
LucaMarconato, line 269: I'd remove the 2 tests test_toggle_3d_points_checkbox and test_toggle_2_5d_shapes_checkbox, and instead just keep the 3rd test, which implicitly tests the first.

src/napari_spatialdata/utils/_utils.py
asarigun: Good point — renamed `yx` to `coords` in both the signature and the docstring, and updated the call site in `_viewer.py` accordingly.

src/napari_spatialdata/utils/_utils.py
asarigun: Done — the docstring now states the axis order explicitly: `(N, 2)` in `(y, x)` order or `(N, 3)` in `(z, y, x)` order.

src/napari_spatialdata/utils/_utils.py
asarigun: Replaced with an explicit `if ndim not in (2, 3): raise ValueError(...)` so `python -O` doesn't strip the check.

src/napari_spatialdata/_sdata_widgets.py
asarigun: Thanks for the careful breakdown — addressed each sub-point:

1. The "3D Settings" label and the two checkboxes are now only created/shown when at least one element across the loaded `SpatialData` objects has a `z` axis. The detection lives in `SdataWidget._sdatas_have_z_axis(...)`, which delegates to a new `SpatialDataViewer._has_z_axis(element)` static method that handles both raster (`DataArray` / `DataTree`) and vector (`DaskDataFrame` / `GeoDataFrame`) elements via `get_axes_names`. 2D-only users no longer see the toggles at all.

2. Renamed both checkboxes to `"Discard z for 3D points"` and `"Discard z for 2.5D shapes"` per your suggestion, since the 2D vs 3D view itself is napari's responsibility — the only thing the toggle controls is whether to drop z. The bindings (and tooltips) were flipped accordingly: a checked box now means "discard z on new layers", and both default to checked because `PROJECT_3D_POINTS_TO_2D` / `PROJECT_2_5D_SHAPES_TO_2D` default to `True`. The `TestUIToggle` test was updated to match the inverted semantics.

3. Both tooltips now explicitly say "Only applies to new layers; layers already displayed are not affected." to make the new-vs-existing scope clear.

src/napari_spatialdata/_sdata_widgets.py
asarigun: Agreed — fully removed the z-range slider from this PR. That means the `QDoubleRangeSlider` widget, the `z_range_header_widget`, `_update_z_slider`, `_on_z_range_changed`, `SpatialDataViewer.get_z_range`, `filter_layers_by_z_range`, the related `superqt`/`QHBoxLayout` imports, and the `TestZBinning` class in the test module are all gone. The PR is now focused on the 3D-viz extension for circles/shapes/points plus the more accessible 3D settings, as you suggested. Happy to revisit z-slicing in a follow-up PR (or a separate plugin / upstream napari) once the value-slider regression is fixed.

src/napari_spatialdata/_sdata_widgets.py
asarigun: Obsoleted by the slider removal — the offending branch no longer exists. If/when we revive the slider in a follow-up, I'll use a fraction of the z-range instead of a fixed `+1.0`.

src/napari_spatialdata/_viewer.py
asarigun: Added a docstring to `_get_transform` that lays out the z-inclusion rule:

> The z axis is included in the returned affine when **both** `include_z` is truthy, **and** the element (and therefore its underlying transformation) has a `z` axis, as reported by `spatialdata.models.get_axes_names`. If `include_z` is requested but the element / transformation does not expose a `z` axis, the flag is silently ignored and a 2D `(y, x)` affine is returned.

Plus the parameters/returns sections. Let me know if you'd rather phrase the transformation part differently.

src/napari_spatialdata/_viewer.py
asarigun: Wired this up — `SpatialDataViewer._has_z_axis(element)` is now a thin per-element check (returns True iff `"z"` is in `get_axes_names(element)`, after a type guard for raster/vector elements). `SdataWidget._sdatas_have_z_axis(sdatas)` iterates over every element across the passed `SpatialData` objects via `sdata._gen_elements()` and short-circuits on the first hit; its result drives the visibility of the "3D Settings" label and the two checkboxes (see also point 1 of the checkbox thread).

src/napari_spatialdata/_viewer.py
asarigun: Resolved by removing the z-slider feature entirely — no more `.compute()` on the dask points just to extract the z range.

src/napari_spatialdata/_viewer.py
asarigun: Great pointer — noting this down for a follow-up. Neuroglancer's per-point shaders are a much more scalable design than ad-hoc Python filtering, and reaching out to the napari team about exposing something similar sounds like the right path.

tests/test_3d_visualization.py
asarigun: Removed `TestZBinning` along with the slider feature itself.

tests/test_3d_visualization.py
asarigun: Collapsed all three pairs into single parametrized tests:

- `Test3DPointsVisualization.test_3d_points_visualization`
- `Test2_5DShapesVisualization.test_2_5d_shapes_visualization`
- `Test2_5DCirclesVisualization.test_2_5d_circles_visualization`

Each is parametrized over `(project_to_2d, expected_ndim) ∈ {(True, 2), (False, 3)}` with `ids=["projected_to_2d", "full_3d"]` so the IDs remain self-documenting.

tests/test_3d_visualization.py
asarigun: Merged into a single `test_affine_transform_preserves_dimensionality` parametrized over `(project_to_2d, expected_data_ndim, expected_affine_shape)`. The identity check uses `np.eye(expected_affine_shape[0])`, and the post-switch "non-identity" assertion is gated on `not project_to_2d` since in the projected-to-2D case the scaled CS still yields a 3x3 identity-like result.

tests/test_3d_visualization.py
asarigun: Both classes are now single parametrized tests (`test_save_points_z_handling` / `test_save_shapes_z_handling`) over `(project_to_2d, expected_data_ndim, z_in_axes) ∈ {(False, 3, True), (True, 2, False)}`, with the z-equality assertion guarded by `if z_in_axes:`.

tests/test_3d_visualization.py
asarigun: Removed both `test_toggle_3d_points_checkbox` and `test_toggle_2_5d_shapes_checkbox`; only `test_toggle_affects_loaded_points` remains, with its docstring updated to note that it implicitly exercises the checkbox <-> config sync.

asarigun:
Thanks again for the thorough review @LucaMarconato — all 15 threads addressed in `a2c765d`.

Highlights of this update:
- Checkboxes renamed to "Discard z for ..." with inverted semantics and a "new layers only" note in the tooltips.
- Visibility of the 3D Settings is now gated on `SdataWidget._sdatas_have_z_axis(...)`, so 2D-only users don't see them at all.
- The z-range slider, `get_z_range`, `filter_layers_by_z_range`, the `TestZBinning` class, and the `superqt` import were all removed; happy to revisit as a follow-up (separate plugin / upstream napari).
- Test classes collapsed via `pytest.parametrize` per your suggestion.
- `_get_transform` now has a docstring explicitly stating that z is included only when `include_z` is truthy **and** the element / transformation has a z axis.

Ready for another look when you have time.

LucaMarconato:
Thanks for the implementing the suggested changes, ready to merge! 😊

PR #395, opened 2026-05-15
Select channels

LucaMarconato:
Replaces #365 

See PR description from @melonora there.


AlbertoFabbri93:
Just to confirm: the goal here is to let users selectively toggle which image channels are visible (e.g., showing DAPI and CD69 while hiding CD45), right?

